PR #5, opened 2016-05-30
Attempt to stand up a style check system within travis-ci.

### LINKED ISSUE #4: Extend travis-ci to reject PR if clang-format fails
We should be able to copy the basic concepts from https://github.com/LLNL/spack.  They use flake8 on python code, but the travis.yml code should be very similar (see https://github.com/LLNL/spack/blob/develop/.travis.yml)

- [x] Install clang-format into the build environment.
- [x] The travis script environment should also run a new custom script: regression/check_format.sh
- [x] The new script should be modeled after: https://github.com/LLNL/spack/blob/develop/share/spack/qa/run-flake8

KineticTheory:
The goal of this change is to add an extra check for PRs that compares modified
code against the .clang-format rule set.  If nonconformances are found, these are
printed as a diff and the style check result is 'fail.'  Users can run
`regression/check_style.sh` manually from their work directory to update code to
match the style requirements (_git-clang-format_ must be installed locally for
this to work).

+ Add a new shell script, check_style.sh, that will run git-clang-format.
+ Remove some of the rules from .clang-format.
+ Update .travis.yml so that clang-format-3.8 is installed (hopefully) and add
  STYLE as a matrix option. Also, run the check_style.sh script after the
  coverate test is complete.
+ fixes #4


src/ds++/DracoMath.hh
```diff
@@ -6,13 +6,8 @@
  * \brief  New or overloaded cmath or cmath-like functions.
  * \note   Copyright (C) 2016 Los Alamos National Security, LLC.
  *         All rights reserved.
- *
- * History:
- * 2013-10-17 Many small header files were combined to create DracoMath.hh.
  */
 //---------------------------------------------------------------------------//
-// $Id$
```
junghans: Can be done in git: https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#Keyword-Expansion

.travis.yml
```diff
@@ -1,11 +1,13 @@
 language: cpp
 
-sudo: false
+sudo: required
+dist: precise
```
junghans: Do you really need sudo and precise? It will make the time until the tests run much longer as travis as more `sudo: false` containers.

.travis.yml
```diff
@@ -30,18 +33,21 @@ env:
...
 
 script:
     # -Wl,--no-as-needed is a workaround for bugs.debian.org/457284
-  - mkdir build && cd build &&
+  - if [[ ${STYLE} ]]; then regression/check_style.sh -t; else mkdir build && cd build &&
     ${HOME}/cmake-${CMAKE_VERSION}/bin/cmake -DRANDOM123_INCLUDE_DIR=${HOME}/Random123-${RANDOM123_VER}/include -DCMAKE_EXE_LINKER_FLAGS="-Wl,--no-as-needed" .. &&
     make -j${J} VERBOSE=1 &&
-    make test ARGS="-E \(c4_tstOMP_2\|viz_tstEnsight_Translator\|cdi_ipcress_tIpcress_Interpreter_Al_BeCu_ipcress\|diagnostics_tDracoInfo\|diagnostics_tDracoInfo_version\|diagnostics_tDracoInfo_brief\|parser_driver4tstConsole_Token_Stream\|c4_tstOMP_1\)" &&
-    make install DESTDIR="${HOME}" && cd -
+    ${HOME}/cmake-${CMAKE_VERSION}/bin/ctest -j ${J} -E \(c4_tstOMP_2\|viz_tstEnsight_Translator\|cdi_ipcress_tIpcress_Interpreter_Al_BeCu_ipcress\|diagnostics_tDracoInfo\|diagnostics_tDracoInfo_version\|diagnostics_tDracoInfo_brief\|parser_driver4tstConsole_Token_Stream\|c4_tstOMP_1\) &&
```
junghans: Is there an advantage of running ctest directly?

regression/check_style.sh
```diff
@@ -0,0 +1,142 @@
...
+# Enable job control
+set -m
+
+##---------------------------------------------------------------------------##
+## Support functions
+##---------------------------------------------------------------------------##
+print_use()
+{
+    echo " "
+    echo "Usage: `basename $0` -d -n -t"
```
junghans: In bash you can use `${0##*/}` instead of `basename $0`.

regression/check_style.sh
```diff
@@ -0,0 +1,142 @@
...
+    echo "     with exit code = 1."
+    echo " "
+}
+
+##---------------------------------------------------------------------------##
+## Sanity Checks
+##---------------------------------------------------------------------------##
+
+# clang-format must be in the PATH
+if ! test "${CLANG_FORMAT_VER}x" = "x"; then
```
junghans: in bash use:
```
if [[ ! ${CLANG_FORMAT_VER} ]]; then
```

regression/check_style.sh
```diff
@@ -0,0 +1,142 @@
...
+  cfver="-${CLANG_FORMAT_VER}"
+else
+  cfver=""
+fi
```
junghans: or use:
```
cfver="${CLANG_FORMAT_VER:+-}${CLANG_FORMAT_VER:+-}"
```

regression/check_style.sh
```diff
@@ -0,0 +1,142 @@
...
+gcf=`which git-clang-format${cfver}`
+cf=`which clang-format${cfver}`
+if test "${gcf}notset" = "notset"; then
```
junghans: Same here `[[ ! ${gcf} ]]` is more bash.

regression/check_style.sh
```diff
@@ -0,0 +1,142 @@
...
+## Check mode
+##---------------------------------------------------------------------------##
+
+if test "${pct_mode}" = "1"; then
+
+  # don't actually modify the files (compare to branch 'develop')
+  cmd='${gcf} --binary ${cf} -f --diff --extensions hh,cc develop'
+  echo $cmd
+  result=`eval $cmd`
+  allok=`echo $result | grep "did not modify" | wc -l`
```
junghans: Use `grep -c` instead of `grep | wc -l`, saves you a subshell.

PR #10, opened 2016-06-27
Add a code style check to travisCI

KineticTheory:
This is WIP...
- Relax some of the checks set by .clang-format.
- Add a new 'style' test to the travisCI matrix.
- The new test uses the provided shell script check_style.sh to ensure new or modified lines of code conform to the style specified by .clang-format.
- Load clang-format-3.8 in the travisCI environment.
- There are several changes recommended by junghans that need to be implemented before this PR is merged. These updates will be added in a separate commit based on the comments that were inadvertantly purged from github when the repo was replaced, but are in a set of emails sent to KT.

To do (comments from junghans)

- [ ] Do you really need sudo and precise? It will make the time until the tests run much longer as travis as more sudo: false containers.
- [x] Is there an advantage of running ctest directly 
  - yes, direct control over `-j`, targets, control parameters, etc.
- [ ] In bash you can use `${0##*/}` instead of `basename $0`.
- [ ] use `if [[ ! ${CLANG_FORMAT_VER} ]]; then` in place of `if ! test "${CLANG_FORMAT_VER}x" = "x"; then`
- [ ] Use `cfver="${CLANG_FORMAT_VER:+-}${CLANG_FORMAT_VER:+-}"` in place of `if ! test "${CLANG_FORMAT_VER}x" = "x"; then ...`
- [ ] Use `grep -c` instead of `grep | wc -l`, saves you a subshell.
- [ ] Use `[[ ! ${gcf} ]]` in place of `if test "${gcf}notset" = "notset"; then`

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -16,12 +16,10 @@ AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: true
 BinPackArguments: false
 BinPackParameters: false
-BreakBeforeBinaryOperators: NonAssignment
 BreakBeforeBraces: Allman
 BreakBeforeTernaryOperators: true
 BreakConstructorInitializersBeforeComma: false
 ColumnLimit:     80
-CommentPragmas:  '^ IWYU pragma:'
 ConstructorInitializerAllOnOneLineOrOnePerLine: true
 ConstructorInitializerIndentWidth: 4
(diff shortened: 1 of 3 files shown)


KineticTheory:
This is not ready yet...

KineticTheory:
@lowrie please review

PR #11, opened 2016-06-28
In the regression system, make github the default location for draco

KineticTheory:
- For all draco regressions, use github by default. There is no need to specify  '-g' when calling regression-master.sh.  Jayenne and Capsaicin still default to svn.
- Change the directory name for the build directory created by the regression system when source code is checked out from github. The top level directory now uses the following pattern (e.g.: `Nightly_gcc-develop`): `<dashboard_type>_<compiler>-<git branch>`
- Teach the regression system to find the draco installation directory when building Jayenne or Capsaicin. This directory now has '-develop' appended (the git branch name).

### DIFF
--- a/regression/ccscs-regress.msub
+++ b/regression/ccscs-regress.msub
@@ -193,7 +193,7 @@ run "module list"
 
 # Use a unique regression folder for each github branch
 if test ${USE_GITHUB} == 1; then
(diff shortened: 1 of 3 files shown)


alexrlongne:
This looks good. LGTM.

PR #24, opened 2016-07-26
Added a simple .vimrc file to conform to style guide

alexrlongne:
This .vimrc file sets tabs to spaces, uses the appropriate number of
spaces and automatically removes trailing white spaces when a file is
saved. Users can copy it to their home directory or source it in their
.vimrc file.


KineticTheory:
LGTM 

In the future, we will want to add a few more style options (and maybe platform specific versions). In particular, I would like to see a hotkey for running clang-format on machines that provide clang-format.

Issue #19, opened 2016-07-07
Use new features of CMake-3.6.0


KineticTheory:
New features:
- [x] Begin supporting OpenBLAS (blas+lapack)
- [x] Take advantage of FindPkgConfig's ability to generate imported targets.
- [x] A "VS_STARTUP_PROJECT" directory property was introduced to specify for Visual Studio Generators the default startup project for generated solutions (".sln" files).
- [x] The "install()" command learned a new "EXCLUDE_FROM_ALL" option to leave installation rules out of the default installation.
- [x] The "list()" command gained a "FILTER" sub-command to filter list elements by regular expression.

Deprecated features (that might cause build issues):
- [x] The "find_library()", "find_path()", and "find_file()" commands no   longer search in installation prefixes derived from the "PATH"  environment variable on non-Windows platforms.  This behavior was  added in CMake 3.3 to support Windows hosts but has proven problematic on UNIX hosts. Users that keep some "<prefix>/bin" directories in the "PATH" just for their tools do not necessarily want any supporting "<prefix>/lib" directories searched.  One may set the "CMAKE_PREFIX_PATH" environment variable with a ;-list of prefixes that are to be searched.
- [x] The precompiled OS X binary provided on "cmake.org" now requires OS X 10.7 or newer.

Reference: https://cmake.org/cmake/help/v3.6/release/3.6.html

PR #29, opened 2016-08-01
Add support for OpenBLAS to satisfy optional LAPACK requirement.

KineticTheory:
WIP 

 * [x] Install OpenBlas on ccscs7 and test.


jhchang-lanl:
Was run on the HPC machine with this case?

KineticTheory:
@jhchang-lanl This is ready for merging.

OpenBLAS doesn't compile with Intel right now so I tested on ccscs7 instead of ML.   I am also using this on my laptop at home.  BTW - CMake didn't learn about OpenBLAS until vesion 3.6.0 (ref #19) so you may need to manually load the cmake/3.6.0 module to test this feature.

PR #31, opened 2016-08-02
Provide better formatting for author list

KineticTheory:
* Use a std::multimap<unsigned,std::string> to store and sort author names. 
* Use a formatted print to wrap at specified column without breaking author names across lines.  Also insert/print appropriate delimeters.


KineticTheory:
* [ ] Regression on ML
* [ ] Style check list
* [ ] Fix travis failure

KineticTheory:
@jhchang-lanl From ml-fey2, I ran `$HOME/draco/regression/regression-master.sh -d Experimental -b Release -p draco -fpr31 &> $HOME/logs/pr31.log &`. Results are on CDash: all green

PR #39, opened 2016-08-12
Fix a couple of recently discovered issues with the regression system.

KineticTheory:
+ Recent changes uncovered a couple of older issues in the regression system.
+ Because `USE_CUDA` now defaults to `OFF`, we need an extra check in `setupCuda` to turn it on in Jayenne if it is turned on in the version of Draco that we are linking to.
+ Fix a problem that could occur if `USE_GITHUB` was never set.
+ Load a newer version of git before trying to rebase from within the nightly regression system.


KineticTheory:
@jhchang-lanl This is ready for merging.

PR #43, opened 2016-08-23
Updates to the regression and script system to support Jayenne at gitlab

KineticTheory:
First cut at updating the regression system to pull jayenne from gitlab.  This should fix regressions on ccscs7, Moonlight and Trinitite.  I will use a 2nd iteration to patch the regression system for darwin and win32.
    
* Provide additional documention about proxy variables found in the draco environment.
* Update sample crontab for ccscs-net servers.  Show commands for integrating gitlab and github repositories with redmine.  We must have a local checkout to allow redmine integration.  These directories are stored at `/ccs/codes/radtran/git.`
* Update the regression system to respect multiple values provided to option `-f`. These are featurebranch values like pr numbers.  By default these are set to `develop`.
* Provide extra logic to set the compiler version as part of the build name shown on cdash.
* Teach the ctest scripts to clone jayenne from gitlab.
* Teach the regression system about jayenne pull requests by updating the helper script fetch_co.sh to take an extra argument that is used to differentiate between github and gitlab and by updating the logic in `draco_regression_macros` to pass the new argument to `fetch_co.sh.`
* This implementation supports running the regression in the following modes:
    
|      mode  |     description  |
|      ------- |   ------------------- |
|      normal |    test draco 'develop' branch w/o jayenne |
|      normal  |   both draco and jayenne are branch 'develop' |
|      draco pr |  test draco 'pr' branch w/o jayenne |
|      jayenne pr | test jayenne 'pr' branch linking to draco develop |
    
* Modes that are not yet supported:
  * `draco pr + jayenne pr` --> Ensure draco pr is good and merged before jayenne pr is reviewed.
  * `jayenne develop against draco pr.` --> might be good to have.
* Cosmetic changes to a few files.

### DIFF
--- a/environment/bin/bash_functions.sh
+++ b/environment/bin/bash_functions.sh
@@ -250,6 +250,7 @@ function rm_from_path ()
 
 ##---------------------------------------------------------------------------##
 ## Toggle LANL proxies on/off
+## https://wiki.archlinux.org/index.php/proxy_settings
 ##---------------------------------------------------------------------------##
 function proxy()
 {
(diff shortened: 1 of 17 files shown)


jhchang-lanl:
There are tests failing for experimental builds of Jayenne. Is this due to this pull request?

KineticTheory:
@jhchang-lanl The tests that are failing are for reasons other than this pr. If this pr was breaking things, all tests would fail.

PR #48, opened 2016-08-28
Restore file that is still needed by Jayenne.

KineticTheory:
* This file is still used by Jayenne. I was overly aggressive when retiring accumulatev.


KineticTheory:
@jhchang-lanl BTW - I expect this file to be re-retired after Rob gets back from vacation.  He is nearly done with some work that will eliminate TigsTrace.

PR #58, opened 2016-09-12
New formatting rules via clang-format

KineticTheory:
The Jayenne team would like to propose a new set of formatting rules to automatically format C++ code used by Draco and Jayenne (and possibly Capsaicin).  We would like to adopt the [style guide from the LLVM coding conventions](http://llvm.org/docs/CodingStandards.html#style-issues).  Most of the low level requirements are captured by the [clang-format](http://llvm.org/releases/3.8.0/tools/clang/docs/ClangFormat.html) tool.  Enforcing this standard will improve readability of code because the formatting and style will be more consistent between files.  Adopting this standard will also reduce the frequency of parsing formatting changes when merging pull requests.

This tool can be run from ccs-net servers using the command `clang-format -i <myfile>`or from Emacs via highlighting a region and pressing F12.  Matt has a local implementation for formatting from within vi that can be provided upon request.

This commit contains the changes to `.clang-format`, and all source files have been updated to meet the demands of the new format. Please review the updated files from this PR to ensure that the new format is reasonable for your team.


src/RTT_Format_Reader/CellDefs.cc
```diff
@@ -84,75 +79,67 @@ void CellDefs::readEndKeyword(ifstream & meshfile)
...
-	   "Error in supplied cell redefinition side types data.");
-    Insist(cell_ordered_sides.size()
-           == static_cast<size_t>(dims.get_ncell_defs()),
-	   "Error in supplied cell redefinition ordered side data.");
+    vector_vector_int const &cell_side_types,
+    std::vector<std::vector<std::vector<size_t>>> const &cell_ordered_sides) {
+  Insist(cell_side_types.size() == static_cast<size_t>(dims.get_ncell_defs()),
+         "Error in supplied cell redefinition side types data.");
+  Insist(cell_ordered_sides.size() ==
+             static_cast<size_t>(dims.get_ncell_defs()),
```
clevelam: This indent seem strange, shouldn't it be lined up with the first open parenthesis?

KineticTheory: The current rule set will align at the opening parenthesis if we had a list of arguments separated by commas.  In this case the argument is an operation that will be indented on the continuation line.  Think of it as

```cpp
cell_ordered_sides = 
    Foo.bar();
```

Without the indent for the continuation line, the above wouldn't look right.

```cpp
cell_ordered_sides = 
Foo.bar();
```

alexrlongne: I can see why the continuation matches opening parenthesis but why does the equality operator cause the next line to be so far to the right?

KineticTheory: The indentation rule is applied to the first argument to `Insist `in exactly the same way as it would be applied to the argument w/o `Insist`.  This is way I was trying to say above. If this still isn't clear, we can talk about it tomorrow.

lowrie:
That's a great way to look at it, KT​.  The first 2 Insists seem to be
inconsistent with the last, but the difference is that they're part of
argument lists, so I guess I'm OK with it.

KineticTheory:
I will merge this even though Travis CI reports a failure.  The failure is that the file templates used by don't conform to the new style guide -- but these files are not valid c++ until they are processed so the style check should not be counted against them.

PR #59, opened 2016-09-13
First cut at configuring the codecov-io feature on GitHub.

KineticTheory:
Update travis CI instructions to improve the coverage checks.
+ Install numdiff by downloading source and building it.
+ Require sudo access so that numdiff can be installed into /usr/bin.
+ Move instructions for installing numdiff, cmake and Random123 to a new script (`.travis-build-dependencies.sh`) that is invoked by travis via the install block.
+ Eliminate the basic build from the matrix. The remaining builds are `STYLE=ON`, `WERROR=ON`, and `COVERAGE=ON`.
+ For `COVERAGE=ON` and `WERROR=ON` builds, we append the compiler variable with the options `--coverage` and `-Werror`, respectively.  In order to avoid using these flags when building numdiff, move these assignments from the before_install block to the before_script block.
+ Include all tests except for `c4_tstOMP_2`.  We exclude this one test because it requires more processors than travis CI can provide to us.
+ When checking out the code use a git depth=3 instead of the default 50.
+ Document the travis environment as a comment block.
+ Update vendor_libraries.cmake so that it can find BLAS and LAPACK located at /usr/lib. These libraries do not have a cmake package file that we normally read to help the build system find these libraries. This requires setting `LAPACK_LIB_DIR` and `BLAS_blas_LIBRARY` in the build environment.
+ Begin using cmake-3.6.2 in place of 3.5.2. Set local environment variables CMAKE and CTEST so they point to the the 3.6.2 version of cmake and ctest.
+ Provide a local environment variable, topdir, that includes the path to the draco sources.  This is different than HOME.
+ Set VENDOR_DIR to avoid a warning from the build system. This value is not used by the build when running under travis-CI.
+ Set the environment variable `RANDOM123_INC_DIR` so that it doesn't need to be set on the cmake configuration line.
+ Remove a couple of deprecated files from environment/Modules.
+ Provide `.codecov.yml` with standard settings and attempt to exclude all code found in test subdirectories.

+ [x]  Attempt to enable tests that are currently disabled in `.travis.yml`:  `ctest -E (c4_tstOMP_2\|viz_tstEnsight_Translator\|cdi_ipcress_tIpcress_Interpreter_Al_BeCu_ipcress\|diagnostics_tDracoInfo\|diagnostics_tDracoInfo_version\|diagnostics_tDracoInfo_brief\|parser_driver4tstConsole_Token_Stream\|c4_tstOMP_1\)`
+ [x] Help the travis build find lapack and run the lapack_wrap tests.

Ensure that we didn't break the required builds (ensure lapack is found and tests are run):
+ [x] Test on ccs-net server
+ [x] Test on ML
+ [x] Test on TT - fails the checks for a new lapack. Fixed with f847d6b.


KineticTheory:
@jhchang-lanl @alexrlongne This is ready for review and merge.

PR #61, opened 2016-09-14
Global key bindings for clang-format through emacs (from KT).

RyanWollaeger:
+ Add lines to draco-global-keys.el to use clang-format with
  f12 key binding for Windows and C-M-tab binding for Mac


KineticTheory:
I can't test the C-M-tab binding on my machine, so I'm trying that you tested that syntax.

RyanWollaeger:
On my Mac, this binding has been tested with success in emacs (on several code regions) for two different .clang-format files (including the llvm version).

PR #63, opened 2016-09-20
Python driver for application unit tests

alexrlongne:
This is my first take at a python driver for unit tests to replace the driver that defined CMake macros, which was slow in some cases. The python file `application_unit_test.py` defines a class that uses the same call as the CMake based script to define and run the application unit test. The class has a function that uses `numdiff` to compare gold files and output files and a function that compares any two files with any specified diff program. It can also search output files for regular expressions. The UnitTest class replaced the old CMake scripts for all of the draco application unit tests (about 8 tests). If needed, I can add more functionality for other code projects when we're ready to try this on harder tests.

* [x] Review by Jae
* [x] Tested in serrano/anaheim
* [x] Suppress valgrind warnings emanating from libpython.


KineticTheory:
@alexrlongne Can you keep both the cmake and the python implementations active at the same time?  Can you move your python versions of AUT functions to `funciton_py`

config/application_unit_test.py
```diff
@@ -0,0 +1,601 @@
...
+# file   config/application_unit_test.py
+# author Alex Long <along@lanl.gov>
+# date   Monday, August 12, 2016, 5:44 pm
+# brief  Provide a python class that aids in creating unit tests that run
+#        interactive user codes (i.e.: run a binary that reads an
+#        input file and diff the resulting output file).
+# note   Copyright (C) 2016, Los Alamos National Security, LLC.
+#        All rights reserved.
+#------------------------------------------------------------------------------#
+# $Id: CMakeLists.txt 6732 2012-09-05 22:28:18Z kellyt $
```
KineticTheory: Remove all of these $Id$ blocks

alexrlongne:
I've added a check in ApplicationUnitTest.cmake that sets the correct driver to either python or cmake based on the extension of the driver file. All tests pass in my checkout of Jayenne with this update. The conversion from cmake to python application unit tests can be done incrementally in Jayenne or Capsaicin.

config/ApplicationUnitTest.cmake
```diff
@@ -243,27 +243,63 @@ macro( aut_register_test )
...
+  if ($PYTHON_TEST EQUAL -1)
+    set($PYTHON_TEST FALSE)
+  else ()
+    set($PYTHON_TEST TRUE)
+  endif ()
+
+
+  # Add python version if python driver file is specified
+
+  if ($PYTHON_TEST)
```
KineticTheory: * Does this syntax work in CMake now?  I though you always needed the curly braces: `${PYTHON_TEST}`.
* I'm pretty sure that you don't use the `$` when setting a variable (line 250, 252).

config/ApplicationUnitTest.cmake
```diff
@@ -243,27 +243,63 @@ macro( aut_register_test )
...
+      -DnumPE=${numPE}
+      -D MPIEXEC=${MPIEXEC}
+      -DMPI_CORES_PER_CPU=${MPI_CORES_PER_CPU}
+      -DSITENAME=${SITENAME}
+      -DPROJECT_BINARY_DIR=${PROJECT_BINARY_DIR}
+      -DPROJECT_SOURCE_DIR=${PROJECT_SOURCE_DIR}
+      ${BUILDENV}
+      -P ${aut_DRIVER}
+      )
+  endif()
```
KineticTheory: Since most of these options are identical, we can store them in a common variable to reduce code duplication.

PR #67, opened 2016-09-23
Add a version check for clang-format

KineticTheory:
Add a version check to the pre-commit hook to ensure clang-format is new enough. Older clang-format version will not apply the rules correctly.


KineticTheory:
@jhchang-lanl @alexrlongne This is ready to be merged.

PR #81, opened 2016-10-12
Reduce memory reserved for aprun/mapn to support KNL.

KineticTheory:
* When packing multiple tests onto a Cray node, we must reserve memory for each PE.  I originally set this to 2 GB.  This value is too high for the KNL partition and would oversubscribe the node's memory. This PR reduces the value to 1.4 GB/PE.
* This commit also removes on obsolete check that ensures that MKL is new enough for the selected Intel compiler.  Newer modulefiles for Intel compilers automatically select the MKL version so this check is no longer needed. 
  * [x] Travis CI checks pass
  * [x] Toss2 checks pass
  * [x] Trinitite checks pass

### DIFF
--- a/config/setupMPI.cmake
+++ b/config/setupMPI.cmake
@@ -271,10 +271,10 @@ macro( setupCrayMPI )
   query_topology()
 
   # Extra flags for OpenMP + MPI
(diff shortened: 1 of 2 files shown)


KineticTheory:
Completed code review with Alex.

PR #111, opened 2016-11-10
Update Travis CI runner scripts

KineticTheory:
* Change some of the tools and setup used by travis (CI system for Draco). I am addressing two issues with this change: (1) We are using clang-format-3.9 on our local systems to check formatting requirements.  We should use the same version when running the checks under Travis, and (2) Recent changes to c4 require OpenMPI 1.10+ but Ubuntu 12 and 14 only know about version 1.6.
+ Modify the script that provides vendors to the build environment on travis to provide clang-format-3.9 and git-clang-format-3.9.  Travis's Ubuntu environment doesn't allow installing these versions via apt-get so do it via this script. Also download, build and install openmpi-1.10.3.  Building MPI takes about 10 minutes, so this will make the checks run much slower.
+ Change the travis environment from Ubuntu 12 to 14 (Trusty).
+ No longer use Travis to setup/install openmpi or clang-format.
+ No longer set some environment variables that were not used.
  * [x] Travis CI checks pass

### DIFF
--- a/.travis-install-dependencies.sh
+++ b/.travis-install-dependencies.sh
@@ -11,7 +11,7 @@
 # not provided by travis or by apt-get.
 
 # preliminaries and environment
-
+set -e
 source regression/scripts/common.sh
 
 topdir=`pwd` # /home/travis/build/losalamos/Draco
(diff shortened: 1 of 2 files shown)


lowrie:
Great!  I'd be happy to merge this, but it's greyed out.

KineticTheory:
@lowrie We don't trust you to merge anything.  JK - I'll update the site to allow you to merge PRs.

PR #135, opened 2016-12-13
Add optional context argument to parse_class template function.

kgbudge:
* Purpose of Pull Request
 
    A second signature of the parse_class template function has been added that takes a second argument of type `Class_Parse_Table::Context_Type const &`. This allows a context object of class-dependent type to be passed to the parser which may provide context to change its behavior.
    
    In a similar fashion, a second version of the parse_class_from_table template function has been added that takes a second `Class_Parse_Table::Context_Type &` argument and passes it as an argument to the `Class_Parse_Table` constructor.
    
    A test has been added to exercise this option and ensure that it works. It also illustrates how this capability is used.

  * [x] Travis CI checks pass

### DIFF
--- a/src/parser/Class_Parse_Table.hh
+++ b/src/parser/Class_Parse_Table.hh
@@ -165,6 +165,60 @@ parse_class_from_table(Token_Stream &tokens) {
   }
 }
 
+//---------------------------------------------------------------------------------------//
+/*! Template for helper function that produces a class object.
+ *
+ * \param tokens Token stream from which to parse the user input.
+ *
+ * \param context A context object that controls the behavior of the parser. This is passed
+ * to the constructor for the Class_Parse_Table.
+ *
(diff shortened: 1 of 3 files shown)


KineticTheory:
@kgbudge We are trying really hard to keep the maximum code with < 80 cols for Draco (Capsaicin has different rules).  Please revert the the changes to the separator comment lines and trim your comment blocks to keep them below 80 cols.  

Background: We do not process comment blocks with clang-format because some of our comment blocks require specific spacing, black lines, etc.  For example, we might use ASCII art to draw something and we don't want this drawing re-indented.  Similarly, we might have LaTeX style equations in a doxygen comment block.  These special use cases require us to disable auto-formatting of comment blocks.  This is why all of the auto-checks pass, but this PR is not in strict conformance with the Draco style guide.

PR #140, opened 2016-12-17
Extend the Abstract_Class_Parser class to allow parsing with context.

kgbudge:
  * [x] Travis CI checks pass
  * [ ] win32 build/test.

The original Abstract_Class_Parser assumed the parse functions for parsing particular child classes were all raw functions with the signature SP<Abstract_Class> parse_function(Token_Stream &). To optionally parse with context, the class needed to accomodate parse functions with the signature SP<Abstract_Class> parse_function(Token_Stream &, Context const &) where Context is the context object type.

To accomplish this, the class was modified to take an additional parse functor type parameter. This allows the parse functors to be any type with an operator(Token_Stream &) defined.

The old case is accomodated by adding a registration function that converts a raw function pointer to a std::fun_ptr object and making std::fun_ptr with suitable typing the default template parameter value for the parse functor type.

The new case is accomodated by adding a new helper class, Contextual_Parse_Functor, which is a closure class that binds a call to a get_context() template parameter function to the second argument of a parse function with the desired signature SP<Abstract_Class> parse_function(Token_Stream &, Context const &). Contextual_Parse_Functor with suitable typing then becomes the parse functor type.

The new test, tstAbstract_Class_Contexual_Parser.cc, illustrates how all this works in practice for an abstract parent class with two child classes with context.

Note that the tstAbstract_Class_Parser.cc test is unchanged. This demonstrates backwards compatibility.

I have also trimmed down the Abstract_Class_Parser_Base, which had too much unnecessary machinery, and moved the essential core of it into the Abstract_Class_Parser.i.hh file.


KineticTheory:
@kgbudge I'm reviewing this code.  I'm getting compile errors under MSVC.  I'll try figure this out and let you know how to amend the change set so we can merge this.  Right now:

```
error C2370: 'rtt_parser::abstract_class_parser_keys' : redefinition; different storage class
   E:\Work\draco\src\parser\Abstract_Class_Parser.cc	line 17
```

src/parser/Abstract_Class_Parser.i.hh
```diff
@@ -14,6 +14,36 @@
...
+ * Helper class defining a table of raw strings created by strdup that will be properly
+ * deallocated using free on program termination.
+ */
+class DLL_PUBLIC_parser c_string_vector {
+public:
+  ~c_string_vector();
+  c_string_vector(void) : data(0) { /* empty */
+  }
+  vector<char *> data;
+};
```
KineticTheory: Please make this change:
```diff
-class DLL_PUBLIC_parser c_string_vector {
+class c_string_vector {
 public:
```

src/parser/test/tstAbstract_Class_Contextual_Parser.cc
```diff
@@ -0,0 +1,409 @@
...
+/*!
+ * \file   parser/test/tstAbstract_Class_Contextual_Parser.cc
+ * \author Kent G. Budge
+ * \date   Tue Nov  9 14:34:11 2010
+ * \brief  Test the Abstract_Class_Contextual_Parser template
+ * \note   Copyright (C) 2016 Los Alamos National Security, LLC.
+ *         All rights reserved.
+ */
+//---------------------------------------------------------------------------//
+// $Id$
```
KineticTheory: Please remove the `$Id$` and the following separator.

src/parser/Abstract_Class_Parser.i.hh
```diff
@@ -14,6 +14,36 @@
...
+ * deallocated using free on program termination.
+ */
+class c_string_vector {
+public:
+  ~c_string_vector();
+  c_string_vector(void) : data(0) { /* empty */
+  }
+  vector<char *> data;
+};
+extern c_string_vector abstract_class_parser_keys;
```
KineticTheory: Kent - I forgot to submit this this request last night (it was still open in my web browser this morning).  This one-liner is also required to get pr140 to work on windows. Sorry for the omission.
```diff
-extern c_string_vector abstract_class_parser_keys;
+DLL_PUBLIC_parser extern c_string_vector abstract_class_parser_keys;
```

PR #153, opened 2017-01-25
Retire the SP smart pointer template by making SP an alias for std::s…

kgbudge:
Retire the SP smart pointer template by making SP an alias for `std::shared_ptr`. This has four consequences:

1. `std::shared_ptr` has no comparison operations with raw pointers, unlike the old SP template.
2. `std::shared_ptr` does not have an `Insist` check for a null pointer with the arrow operator.
3. The conversion to bool is explicit for `std::shared_ptr`. If tests and operator! are equivalent to explicit casts to `bool` under C++11, but an explicit cast is required in other contexts. This bites us mostly for functions like `UnitTest::check()` that expect a bool argument, but I have eased this by adding a template version of `check()` that accepts `std::shared_ptr` in place of bool. I have also done the same for `Token_Stream::check_semantics()`.
4. SP handles downcasts (via dynamic_cast) implicitly and throws an error if it was unsuccessful.  `std::shared_ptr` handles dynamic_cast the same manner a raw pointer does

All of the above follow the well-known design criteria "behave **exactly** like a raw pointer but delete the memory you point to on destruction."

  * [x] Travis CI checks pass
  * [x] Tested on Snow
  * [x] Tested on Win32
  * [x] Tested on KNL
  * [ ] Capsaicin ready
  * [x] Jayenne ready
* Refs [#862](https://rtt.lanl.gov/redmine/issues/862)


src/ds++/UnitTest.hh
```diff
@@ -108,6 +109,11 @@ public:
   DLL_PUBLIC_dsxx bool passes(std::string const &passmsg);
   DLL_PUBLIC_dsxx bool check(bool, std::string const &checkmsg,
                              bool fatal = false);
+  template <class T>
```
KineticTheory: I would like to see a doxygen comment for this function.

src/ds++/SP.hh
```diff
@@ -13,321 +13,14 @@
...
- * Note that it is perfectly acceptable to call delete on a NULL pointer.
- */
-template <typename T> void SP<T>::free() {
-  // if the count goes to zero then we free the data
-  if (r_ && --r_->refs == 0) {
-    delete p_;
-    delete r_;
-  }
-}
+using std::dynamic_pointer_cast;
```
KineticTheory: This type of namespace injection makes me nervous.  We should discuss this at the meeting tomorrow.

src/parser/Token_Stream.hh
```diff
@@ -203,6 +204,12 @@ public:
       report_semantic_error(message);
   }
```
KineticTheory: `rtt_parser::check_semantics` needs doxygen documentation.

src/ds++/SP.hh
```diff
@@ -13,321 +13,14 @@
...
- *
- * \param sp_in smart pointer of type SP<T>
- */
-template <typename T>
-SP<T>::SP(const SP<T> &sp_in) : p_(sp_in.p_), r_(sp_in.r_) {
-  // advance the reference to T if not empty
-  if (r_)
-    r_->refs++;
-}
+template <class T> using SP = std::shared_ptr<T>;
```
KineticTheory: Nice!

KineticTheory:
@kgbudge I'm trying to build Jayenne against your PR right now.  Once I have that working, I will merge #153

PR #157, opened 2017-01-28
[WIP] First real attempt to prevent CI scripts from clobbering each other.

KineticTheory:
**Background**

If a PR is submitted and then updated a short time later, the CI system would fall over because it would kick off jobs for the updated PR that would clobber the CI jobs running for the initial PR.

**Resolution**
I have chosen to use the `flock` tool on Linux to prevent the second set of jobs from starting until the first set have finished. There are a few other updates in this PR that provide general cleanup in the CI and regression scripts. The script `sync_repository.sh` was modified by adding file-lock checking and forcing the script to wait until all subprocesses are complete before exiting. Additionally, a second check was put into place (via a new bash function `allow_file_to_age` to ensure that individual builds do not start too soon.

**Other changes in this PR**
+ CI tests will begin running on snow.
+ The `sync_repository.sh` script was updated to correctly load the module environment on snow.
+ I use the _access_ queue to run CI jobs on Toss2 machines. Toss3 machines will use the _dev_ queue (once available). Update script logic to look for and use the _dev_ queue if available for the current user.
+ Updated the `sync_repository.sh` and `update_regression_scripts.sh` scripts to use internal logic to save all output to log files. Crontabs were updated since these scripts no longer need to be started with syntax that captures output.
+ Logic was added to `update_regression_scripts.sh` to identify old log files and old CI build directories. Once I am happy with how this logic works, the script will remove these old files and directories.
+ Modify the bashrc file for the red-fta and redcap machines. If the 'user_contrib' modulefiles are available at $HOME, use those instead of the published files.
+ Alex and I submitted a couple of PRs to try and identify test failures on the KNL nodes of trinitite.  It looks like one of the CFLAGS was changed and never changed back: `-fp-model precise` --> `-fp-model fast`.  The later should be the default except on KNL.
+ refs [#841](https://rtt.lanl.gov/redmine/issues/841)

  * [x] Travis CI checks pass

### DIFF
--- a/config/unix-intel.cmake
+++ b/config/unix-intel.cmake
@@ -41,14 +41,13 @@ if( NOT CXX_FLAGS_INITIALIZED )
   set( CMAKE_C_FLAGS_DEBUG
     "-g -O0 -inline-level=0 -ftrapuv -check=uninit -fp-model precise -fp-speculation safe -DDEBUG")
   set( CMAKE_C_FLAGS_RELEASE
-    "-O3 -fp-speculation fast -fp-model precise -pthread -DNDEBUG" )
+    "-O3 -fp-speculation fast -fp-model fast -pthread -DNDEBUG" )
   # [KT 2017-01-19] On KNL, -fp-model fast changes behavior significantly for
   # IMC. Revert to -fp-model precise.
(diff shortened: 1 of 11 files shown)


KineticTheory:
I'm seeing some funny behavior on sn-fey. I'm marking this as WIP until I can confirm the desired behavior.

PR #160, opened 2017-02-08
Significant update to regression scripts.

KineticTheory:
The initial goal of this work was to prevent two simultaneous CI instances of the same PR.  However, after working through several potential solutions it has resulted in a bigger set of changes as detailed below. refs #841
+ Begin running nightly tests and PRs on snow. Also teach the snow regression scripts to use qos=dev and to use all 36 cores on each node.
+ Use `flock` to prevent some scripts from having more than one concurrent instance. Also add some job control cleanup to the end of these scripts. `flock` doesn't seem to work well with if the log file (on a NFS mount) is used as the locking file.  Instead I create an empty file at /var/tmp. This mechanism now is used in these files:
  + regression-master.sh
  + sync_repository.sh
  + sync_vendors.sh
+ Teach some scripts to capture and write all of their output to a log file. This simplifies the crontab commands that previously redirected all output to a file.  This change was made in these files:
  + sync_vendors.sh
  + sync_repository.sh
  + update_regression_scripts.sh
+ Clean up old work directories and log files automatically (delete files or directories that are more than 14 days old).
  * [ ] Travis CI checks pass
  * [ ] Manually test these scripts on Snow, Moonlight, Trinitite and ccscs7.

### DIFF
--- a/regression/ccscs-crontab
+++ b/regression/ccscs-crontab
@@ -1,16 +1,16 @@
 # crontab for ccscs7
 
 # Keep vendor installations in sync between ccs-net servers.
-45 21 * * 0-6 /scratch/regress/draco/regression/sync_vendors.sh &> /scratch/regress/logs/sync_vendors_ccscs7.log
+45 21 * * 0-6 /scratch/regress/draco/regression/sync_vendors.sh
 
 # Update the regression scripts.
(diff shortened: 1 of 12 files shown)


KineticTheory:
I'm happy with this PR for now.  There are a few more changes I want to add, but they need to go into a different PR.

PR #168, opened 2017-02-24
Add option to UnitTest to suppress pass messages in UnitTest::check

kgbudge:
This new feature addresses the problem of finding the one failing test among a slew of pass messsages for a test code that does many checks. 

By default this option is off, so existing behavior is unchanged. When this option is turned on, messages are generated by UnitTest::check only if the bool argument is false (meaning the test failed.) This can be extremely useful for debugging a test executable that has many check calls in it.

  * [x] Travis CI checks pass

### DIFF
--- a/src/c4/ParallelUnitTest.cc
+++ b/src/c4/ParallelUnitTest.cc
@@ -23,6 +23,8 @@ namespace rtt_c4 {
  * \arg argv A list of strings containg the command line arguments
  * \arg release_ A function pointer to this package's release function.
  * \arg out_ A user specified iostream that defaults to std::cout.
+ * \arg verbose_ flags whether to print messages for successful tests. Defaults
+ * to true.
  * \exception rtt_dsxx::assertion An exception with the message "Success" will
  * be thrown if \c --version is found in the argument list.
  *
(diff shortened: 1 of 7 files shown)


KineticTheory:
@kgbudge 

1. Please update the title of this PR to less than 80 chars so it displays properly.
2. Either delete the boilerplate text in the description ('Purpose of the pull ...') or provide links to an open ticket.

@jhchang-lanl @alexrlongne @warsa @RyanWollaeger  I must confess, I don't understand why this change is needed. It seems that this type of fine-grain control should be handled in the unit test instead of adding complexity to the UnitTest framework.  However, I would like other developers to weigh in on this proposal before any decision is made.

warsa:
I can see why this option might be useful.

RyanWollaeger:
I can see how this would be useful as well, when there are a lot of results per unit test.
And it is nice to have the tests default to verbose.
However, to suppress messages from certain test functions, would this require addition of instantiations of the UnitTest class?

RyanWollaeger:
Could a "set_verbose" function be added to merely update the verbose state?

kgbudge:
At present, the UnitTest is wired one way or another at construction. A set_verbose would allow finer-grained use, but the idea really is to turn off a slew of "passed" messages that you have to hunt through to get the "failed" message. This is a genuine headache.

________________________________

PR #170, opened 2017-02-27
Ensure timely deallocation of parser resources.

kgbudge:
Ensure that destructor of `Class_Parse_Table<Quadrature>` releases the parsed quadrature in a timely manner. Previously the parsed quadrature would not be released until general end of execution cleanups or until another quadrature was parsed.

Provide a way to free the global pointer to the internal unit system. I'm not sure why this is implemented as a raw pointer rather than a `shared_ptr,` but the ability to free this storage manually would be useful even if it were implemented as a` shared_ptr.`

  * [x] Travis CI checks pass

### DIFF
--- a/src/parser/test/tstutilities.cc
+++ b/src/parser/test/tstutilities.cc
@@ -936,6 +936,7 @@ void tstutilities(UnitTest &ut) {
     else
       FAILMSG("did NOT parse bare expression to SI correctly");
     bare_expression.rewind();
+    free_internal_unit_system();
   }
   return;
 }
--- a/src/quadrature/Quadrature__parser.hh
+++ b/src/quadrature/Quadrature__parser.hh
@@ -33,6 +33,8 @@ public:
 
   Class_Parse_Table();
 
+  virtual ~Class_Parse_Table() { child_.reset(); }
+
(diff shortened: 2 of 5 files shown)


src/parser/test/tstutilities.cc
```diff
@@ -936,6 +936,7 @@ void tstutilities(UnitTest &ut) {
     else
       FAILMSG("did NOT parse bare expression to SI correctly");
     bare_expression.rewind();
+    free_internal_unit_system();
```
kgbudge: Exercises the new function to free the storage for the internal unit system.

src/quadrature/Quadrature__parser.hh
```diff
@@ -33,6 +33,8 @@ public:
 
   Class_Parse_Table();
 
+  virtual ~Class_Parse_Table() { child_.reset(); }
+
```
kgbudge: This is the key element of this change. This ensures that child_, a static pointer to the  parsed quadrature, is reset when parsing is complete. Thus the storage will not linger after the client has destroyed his last pointer to the parsed quadrature.

KineticTheory:
LGTM

PR #173, opened 2017-03-02
Use std::expm1() in conversion from Planck to Rosseland.

dholladay00:
* Purpose of Pull Request
  * This pull request changes how Planck2Rosseland factor is calculated, using a std::expm1(x) which is accurate even for small (x), refs #2 (882)

  * [ ] Travis CI checks pass

### DIFF
--- a/src/cdi/CDI.hh
+++ b/src/cdi/CDI.hh
@@ -196,9 +196,14 @@ static double polylog_series_minus_one_planck(double const x,
 static double Planck2Rosseland(double const freq, double const exp_freq) {
   Check(rtt_dsxx::soft_equiv(exp_freq, std::exp(-freq)));
 
-  double const freq_4 = freq * freq * freq * freq;
+  double const freq_3 = freq * freq * freq;
 
-  double const factor = NORM_FACTOR * freq_4 / std::expm1(freq);
+  double factor;
+
+  if (freq > 1.0e-5)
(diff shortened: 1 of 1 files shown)


RyanWollaeger:
Several machines appear to have caught some failed assertions associated with this header:
https://rtt.lanl.gov/cdash/testDetails.php?test=489229&build=856

dholladay00:
Yep, I built Draco and am seeing that `Assertion: rosseland >= 0.0, failed in /users/dholladay00/repos/Draco/src/cdi/CDI.hh, line 747.`

So, it's definitely something wrong with my change.

KineticTheory:
The old if/else logic was in place to avoid division by zero when nu is very close to zero.  That might need to go back in.

dholladay00:
More extensive tests failed due to overflow. This is *probably* occurring when `freq` is large, thus making `expm1(freq)` overflow (that is unverified but my guess without further investigation). This was the wisdom baked in to the previous way of multiplying numerator and denominator by `exp(-freq)`. Even in this form, there is a way to use `expm1` in that `1 - exp(-freq) = -expm1(-freq)`.

thus it would be `exp_freq * freq * freq_3 / -std::expm1(-freq)`

Rolling back all of the changes is an option that we know won't hurt anything, but I'd like to see if the new small frequency approximation breaks anything b/c it should be more accurate.

PR #177, opened 2017-03-09
Add C4 interface for MPI_Sendrecv function.

kgbudge:
I have run into situations in my code where this seems the best operation for what I want to accomplish.

  * [x] Travis CI checks pass

### DIFF
--- a/src/c4/C4_Serial.hh
+++ b/src/c4/C4_Serial.hh
@@ -73,6 +73,14 @@ DLL_PUBLIC_c4 int send_udt(const T * /*buffer*/, int /*size*/,
   return C4_SUCCESS;
 }
 
+template <typename TS, typename TR>
+DLL_PUBLIC_c4 int send_receive(TS * /*sendbuf*/, int /*sendcount*/,
+                               int /*destination*/, TR * /*recvbuf*/,
+                               int /*recvcount*/, int /*source*/,
+                               int /*sendtag*/, int /*recvtag*/) {
+  Insist(false, "send_receive is not support for C4_SCALAR builds.");
+}
+
 //---------------------------------------------------------------------------//
 
 template <class T>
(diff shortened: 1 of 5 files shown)


src/c4/C4_Serial.hh
```diff
@@ -73,6 +73,14 @@ DLL_PUBLIC_c4 int send_udt(const T * /*buffer*/, int /*size*/,
   return C4_SUCCESS;
 }
 
+template <typename TS, typename TR>
+DLL_PUBLIC_c4 int send_receive(TS * /*sendbuf*/, int /*sendcount*/,
+                               int /*destination*/, TR * /*recvbuf*/,
+                               int /*recvcount*/, int /*source*/,
+                               int /*sendtag*/, int /*recvtag*/) {
+  Insist(false, "send_receive is not support for C4_SCALAR builds.");
```
KineticTheory: We talked about this yesterday. I think we could implement a 'scalar' version of this that simply copies the `sendbuf` into the `recvbuf`.  Theoretically, Draco supports a `C4=SCALAR` (no MPI) mode and I would like to make an attempt keep this mode alive.  Can you make this change, removing the fatal error?

KineticTheory:
Trinitite is on a 5-day DST.  I won't require this build to report for this PR.

src/c4/C4_Serial.hh
kgbudge: I looked at this, and the problem is that there are separate tags for the send and receive halves. When they are identical, this is indeed a simple copy. Since I'm not sure how to handle this for tags that are not identical (I'm not sure it *can* be handled in serial) I'm disinclined to half-implement it. What do you think?

KineticTheory: I'm good with your decision.

PR #178, opened 2017-03-14
Compton: Remove an unneeded include and improve test coverage

keadyk:
* This pull request should increase test coverage for Compton.hh to 100%. It also removes an unneeded include from Compton.cc.

  * [x] Travis CI checks pass

### DIFF
--- a/src/compton/Compton.cc
+++ b/src/compton/Compton.cc
@@ -13,7 +13,6 @@
 #include "ds++/Assert.hh"
 // headers provided in Compton include directory:
 #include "compton_file.hh"
-#include "multigroup_compton_data.hh"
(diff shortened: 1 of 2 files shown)


keadyk:
I added an additional commit to take care of the gcc-4.8.5 Release failures on ccscs7.  

In tCompton.cc, I check to see whether an exception is thrown when a non-existent file is passed to the Compton constructor. If DBC is on, a Require() in Compton::Compton throws a rtt_dsxx::assertion, which I was checking for in the original version of the test.

If DBC is off, the CSK library will still throw an exception on the bad file... but it's just an int exception. I added an additional clause to the try-catch statement in tCompton to handle this scenario. The test should now pass regardless of DBC level.

PR #188, opened 2017-03-24
Add capability to extend debug options set.

kgbudge:
New functions added to allow applications to add highly application-specific options to the debug mask.

The immediate need is for Capsaicin to have a debug option to look for cycles in an input mesh. This is specific enough to Capsaicin that I do not want to introduce it as a general debug option. However, I have been thinking about making the debug option set extensible for some time in anticipation that a situation like this would come up.

We may wish to consider dropping some existing "standard" options if there is consensus that they are too application-specific.

  * [x] Travis CI checks pass


KineticTheory:
@kgbudge I'll try to review and merge this today.

src/parser/Debug_Options.cc
```diff
@@ -102,10 +144,87 @@ string debug_options_as_text(unsigned const debug_options) {
...
+  }
 
   return Result;
 }
 
+//---------------------------------------------------------------------------------------//
+/*! Add a new debug option to the debug parser specific to an application. This version
+    assigns the next available bit.
+ *
+ * \param Debug option keyword
```
KineticTheory: Can you use `\param[in] option_name Debug option keyword`? 
* Doxygen expects the variable name to be listed. 
* The [in] out looks nice in the docs. 
* Ditto for remaining function documentation below.

KineticTheory:
@kgbudge There were some issues with the CI testing.  Once you push the minor fixes I requested, a new sets of tests will start, hopefully running to completion this time.

src/parser/Debug_Options.cc
```diff
@@ -12,10 +12,67 @@
 
 #include "Debug_Options.hh"
```
kgbudge: It's been my practice with Class.cc, Class.hh file pairs to have the include directive for Class.hh come first in Class.cc. This ensures that there will be an error when Class.cc is compiled if some include file or using statement that should be in Class.hh (to make it self-contained) is missing.

KineticTheory:
@kgbudge Waiting for Travis to finish its checks before I merge.

PR #199, opened 2017-04-12
Stand up nightly regression for Intel-17.0.2 and OpenMPI-2.0.2.

KineticTheory:
+ On Snow, create two new special builds:
  - newtools - Intel-17.0.2 + OpenMPI-2.0.2
  - gcc610   - GNU-6.1.0 + OpenMPI-2.0.2
+ Fix some spelling errors.
+ Remove or hide some debug output.
+ [Refs #893](https://rtt.lanl.gov/redmine/issues/893)
  * [ ] Travis CI checks pass

### DIFF
--- a/regression/sync_repository.sh
+++ b/regression/sync_repository.sh
@@ -24,6 +24,7 @@
 #    all pull requests.
 
 target="`uname -n | sed -e s/[.].*//`"
+verbose=off
 
 # Locate the directory that this script is located in:
 scriptdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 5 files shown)


regression/sync_repository.sh
```diff
@@ -24,37 +24,33 @@
...
+  echo "looking for locks..."
+  echo "   FLOCKER     = ${FLOCKER}"
+  echo "   lockfile    = $lockfile"
+  echo "   logfile     = $logfile"
+  echo "   script name = ${0}"
+  echo "   script args = $@"
+fi
 
 # Prevent multiple copies of this script from running at the same time:
 [ "${FLOCKER}" != "${lockfile}" ] && exec env FLOCKER="${lockfile}" flock -en "${lockfile}" "${0}" "$@" || :
```
RyanWollaeger: Instead of evaluating to a no-op, would it be useful to set an alternate `verbose` variable?
When `verbose` and the alternate are both "on", something like "another copy of ${0} is running..." could be echoed.
Also, is `verbose` set manually in this script for testing only?

RyanWollaeger: Never mind - I see from the crontab this scenario should almost never happen during testing.

PR #205, opened 2017-05-08
Add prefix sum function to C4 library

alexrlongne:
+ The MPI scan function calculates a prefix sum (often used in determining offsets in parallel problems) without forming a n_proc length array on every node.
+ Templated prefix_sum function was added to the "blocking" section of the C4 library
+ A section for prefix_sum was added to the "tstReduction" test
* Purpose of Pull Request
  * Address Jayenne #892 (https://rtt.lanl.gov/redmine/issues/892)
  * [ ] Travis CI checks pass
  * [x] [Trinitite checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco) (regression broken after slurm install).

### DIFF
--- a/src/c4/C4_Functions.hh
+++ b/src/c4/C4_Functions.hh
@@ -472,8 +472,7 @@ DLL_PUBLIC_c4 std::string get_processor_name();
 /*!
  * \brief Return the value of the prefix sum at this processor.
  *
- * \param node_value
- * Current node's value of variable to be prefix summed
+ * \param node_value Current node's value of variable to be prefix summed
  * \return Sum of value over nodes up to and including this node.
  */
 template <typename T> DLL_PUBLIC_c4 T prefix_sum(T &node_value);
--- a/src/c4/C4_Functions.hh
+++ b/src/c4/C4_Functions.hh
@@ -476,7 +476,7 @@ DLL_PUBLIC_c4 std::string get_processor_name();
  * Current node's value of variable to be prefix summed
  * \return Sum of value over nodes up to and including this node.
  */
-template <typename T> DLL_PUBLIC_c4 T prefix_sum(T& node_value);
+template <typename T> DLL_PUBLIC_c4 T prefix_sum(T &node_value);
 
 } // end namespace rtt_c4
 
--- a/src/c4/C4_Functions.hh
+++ b/src/c4/C4_Functions.hh
@@ -466,6 +466,18 @@ DLL_PUBLIC_c4 bool isScalar();
 //! Return the processor name for each rank.
 DLL_PUBLIC_c4 std::string get_processor_name();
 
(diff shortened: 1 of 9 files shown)


src/c4/test/tstReduction.cc
```diff
@@ -190,11 +191,72 @@ void array_reduction(rtt_dsxx::UnitTest &ut) {
...
+
+  if (!soft_equiv(xdbl_prefix_sum, dbl_answer))
+    ITFAILS;
+
+  if (ut.numFails == 0)
+    PASSMSG("Prefix sum ok.");
+  return;
+}
+
+//---------------------------------------------------------------------------//
```
KineticTheory: Please add tests for uint32_t and float.  These should be trivial to add.

src/c4/C4_MPI.t.hh
```diff
@@ -353,6 +353,15 @@ template <typename T> DLL_PUBLIC_c4 void global_max(T *x, int n) {
...
 
+//---------------------------------------------------------------------------//
+
+template <typename T> DLL_PUBLIC_c4 T prefix_sum(T &node_value) {
+  T prefix_sum = 0;
+  MPI_Scan(&node_value, &prefix_sum, 1, MPI_Traits<T>::element_type(), MPI_SUM,
+           communicator);
+  return prefix_sum;
+}
+
```
KineticTheory: You need to add an equivalent function to C4_Serial.hh.

src/c4/C4_Functions.hh
```diff
@@ -466,6 +466,18 @@ DLL_PUBLIC_c4 bool isScalar();
...
 DLL_PUBLIC_c4 std::string get_processor_name();
 
+//---------------------------------------------------------------------------//
+// prefix_sum
+//---------------------------------------------------------------------------//
+/*!
+ * \brief Return the value of the prefix sum at this processor.
+ *
+ * \param node_value
+ * Current node's value of variable to be prefix summed
```
KineticTheory: Put the description on the same line as the `\param` or check that this produces the correct output (make autodoc).  See examples at https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html.

src/c4/C4_MPI.t.hh
alexrlongne: prefix_sum now has a serial implementation and I tested and updated the C4_SERIAL build (with your help)

src/c4/C4_Functions.hh
alexrlongne: I've moved the description to the same line, as is done in the examples.

src/c4/C4_MPI.t.hh
KineticTheory: Are there any DbC statements that can be inserted into this function?  Is there a valid range for node_value?

PR #227, opened 2017-05-29
Scalable Invert_Comm_Map

lowrie:
* Change Invert_Comm_Map so that it no longer uses num_procs arrays.
  * [Refs #914 (redmine)](https://rtt.lanl.gov/redmine/issues/914)
  * Also requires adding MPI_ANY_SOURCE capability to C4; [refs #936 (redmine)](https://rtt.lanl.gov/redmine/issues/936).  This capability is attained by passing a source argument as rtt_c4::any_source. 
  * Added C4_Status class, which wraps MPI_Status.  This is needed for any_source capability.
  *  At this stage, the any_source  is the minimal capability to get Invert_Comm_Map working.  In particular, C4_Req::wait() and complete() now take the optional argument of an C4_Status pointer. 
  * any_source has not been implemented yet for probe functions, which will need an interface change. 
  * Code cleanup.

  * [x] Travis CI checks pass


src/c4/C4_MPI.cc
```diff
@@ -148,6 +156,8 @@ void blocking_probe(int source, int tag, int &message_size) {
 
 //---------------------------------------------------------------------------//
 void wait_all(int count, C4_Req *requests) {
+  Require(count >= 0);
```
kgbudge: Could the argument type for count be changed to unsigned, to make this precondition check unnecessary?

src/c4/test/tstC4_Status.cc
```diff
@@ -0,0 +1,95 @@
...
+}
+
+//---------------------------------------------------------------------------//
+int main(int argc, char *argv[]) {
+  rtt_c4::ParallelUnitTest ut(argc, argv, rtt_dsxx::release);
+  try {
+    if (rtt_c4::nodes() == 2)
+      tst2Procs(ut);
+  }
+  UT_EPILOG(ut);
```
lowrie: It is run only on 2 procs, via a cmake change, but in case that every changes/fails, I'll put in a fail here.

src/c4/C4_MPI.cc
lowrie: Yes, that's a better fix...I'll try that.

KineticTheory:
Whack-a-mole...
```
/home/travis/build/lanl/Draco/src/c4/C4_MPI.cc:165:23: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   for (int i = 0; i < count; ++i) {
                       ^
cc1plus: all warnings being treated as errors
```

lowrie:
Is this PR OK?

KineticTheory:
@lowrie Alex is taking one more look at your PR.

src/c4/test/tstC4_Status.cc
```diff
@@ -0,0 +1,95 @@
...
+  C4_Req request;
+
+  const int num_int = 2;
+  const int num_double = 5;
+  const int tag = 101;
+
+  if (my_proc == 0) {
+    std::vector<int> send_buffer(num_int);
+    std::vector<double> recv_buffer(num_double);
+    rtt_c4::receive_async(request, &recv_buffer[0], num_double, 1, tag);
```
lowrie: yeah, good idea.  It does check status.get_source(), but this is a further check.

src/c4/Invert_Comm_Map.cc
```diff
@@ -17,74 +17,118 @@
...
+      // we can ignore the request returned, because our send buffers are
+      // not shared, and we'll wait on the receives below.
+      send_async(&(it->second), 1, it->first, tag);
+    }
+  }
+
+  // Wait on the receives and populate the map
+  C4_Status status;
+  for (int i = 0; i < num_recv; ++i) {
+    recvs[i].wait(&status);
```
lowrie: C4 does provide a wait_all, which wraps MPi's wait_all.  But the wrapper ignores the status (it's passed MPI_STATUSES_IGNORE), which I need.  It'll need to be modified to support an array of C4_Status.  I don't need this for Invert_Comm_Map, so I'd rather not go down this route, until it's needed.  Same goes with wait_any().

PR #234, opened 2017-06-08
Update Compton.* to reflect new interface for CSK v0.2.0

keadyk:
* Purpose of Pull Request
  * This PR updates the compton interface class in draco to match the v0.2.0 release of CSK 
  *  refs #938 (Jayenne redmine number)
  *  With the new version of CSK, the draco class can be used to access existing Jayenne or  Capsaicin-style multigroup libraries, build them from scratch using library routines in CSK, and interpolate multigroup opacities and average frequency ratios in electron temperature (again using CSK library routines)
  * CSK v0.2.0 is released on ccs-net for gcc-6.3.0 and clang-4.0.0
  * The draco Compton unit test has also been reduced to a more minimal working example to decrease test time!

  * [x] Travis CI checks pass


keadyk:
I moved the csk-0.2.0 module file out of /scratch/vendors/Modules.lmod, so not as to break the nightlies!

keadyk:
I moved 0.2.0 back into Modules.lmod/csk and added a .modulerc file to set 0.1.0 as the default version. I'll update this to 0.2.0 once this PR is merged (thank you for this suggestion, KT!)

src/compton/Compton.cc
```diff
@@ -57,10 +57,21 @@ Compton::Compton(const std::string &filehandle) {
...
  * \param grp_bds    A vector containing the multigroup bounds (in keV)
+ * \param opac_type  The type of opacity to build. Valid options for CSK v0.2
+ *                   are "jayenne" (for IMC-style opacities) or "capsaicin"
+ *                   (for Sn-style opacities). Any other string will cause CSK 
+ *                   to throw an exception 
+ * \param wt_func    The frequency weighting function used to numerically 
+ *                   integrate the opacities. Valid options for CSK v0.2 are
+ *                   "flat", "wien" or "planck." Any other string will cause 
+ *                   CSK to throw an exception.
+ * \param induced    Bool to toggle consideration of induced effects off/on
```
KineticTheory: * Please provide [in], [out] or [inout] for each param.
* I'm not crazy about using the strings `capsaicin` or `jayenne` for `opac_type`.  Is there a better way to express these two modes?

src/compton/Compton.cc
```diff
@@ -57,10 +57,21 @@ Compton::Compton(const std::string &filehandle) {
...
  * \param n_xi       The number of angular points/Legendre moments desired
  */
 Compton::Compton(const std::string &filehandle,
-                 const std::vector<double> &grp_bds, const size_t nxi) {
+                 const std::vector<double> &grp_bds,
+                 const std::string opac_type, const std::string wt_func,
```
KineticTheory: I think `opac_type` and `wt_func` are of input only.  As strings, they should be of type `const &` to avoid making a copy (an allocation).

src/compton/Compton.cc
```diff
@@ -107,25 +123,47 @@ Compton::Compton(const std::string &filehandle,
  * \param etemp The SCALED electron temperature ( temp / electron rest-mass )
  * \return      n_grp x n_grp x n_xi interpolated scattering kernel values
  */
-std::vector<std::vector<std::vector<double>>>
-Compton::interpolate(const double etemp) {
+std::vector<std::vector<std::vector<std::vector<double>>>>
+Compton::interpolate_csk(const double etemp) {
```
KineticTheory: A 4-D vector seems pretty intense.  We might want to consider a different container since this requires 4 pointer dereferences to access each data element.

src/compton/Compton.hh
```diff
@@ -68,13 +68,22 @@ public:
   explicit Compton(const std::string &filehandle);
 
   //! Constructor to build a multigroup library from an existing pointwise file
+  //! Defaulting the number of angular evals (n_xi) to zero causes the CSK
```
KineticTheory: Move this text the function documentation block in the `.cc` file.  Try to keep this documentation to a single line.

src/compton/Compton.hh
```diff
@@ -68,13 +68,22 @@ public:
...
-  std::vector<std::vector<std::vector<double>>>
-  interpolate(const double etemp) const;
+          const std::string opac_type, const std::string wt_func,
+          const bool induced, const size_t n_xi = 0);
+
+  //! Interpolation of all csk data to a certain electron temperature:
+  std::vector<std::vector<std::vector<std::vector<double>>>>
+  interpolate_csk(const double etemp);
+  std::vector<std::vector<std::vector<std::vector<double>>>>
+  interpolate_csk(const double etemp) const;
```
KineticTheory: Is the only difference the `const`?  Is there any way to delete the non-const flavor?  ditto for the next function.

src/compton/test/tCompton.cc
```diff
@@ -98,6 +103,17 @@ void compton_file_test(rtt_dsxx::UnitTest &ut) {
   if (!soft_equiv(interp_data[0][0][3], 3.59663442e-02))
     ITFAILS;
 
+  if (!soft_equiv(interp_nu_data[0][0], 1.5000000e+00))
```
KineticTheory: If you need to make this check tighter, soft_equiv takes a 3rd argument that is the relative tolerance for the check.

src/compton/Compton.cc
keadyk: Are you talking about the fact that they're strings, or my actual choice of names? I'm definitely open to suggestions!

src/compton/Compton.cc
keadyk: Okay! I'd like to talk with you about the best way to 'flatten' this data before I make any changes :)

src/compton/Compton.cc
KineticTheory: Just change `\param` to `\param[in]`, `\param[out]` or `\param[inout]`. Most of our code doesn't use this newish Doxygen feature, but I'm trying to ensure that new code uses these _intent_ declarations.

src/compton/Compton.hh
keadyk: Yep, that's the only difference. I implemented const- and non-const versions for flexibility's sake, but I'm having a hard time envisioning why anyone downstream would need a non-const Compton object. Let me think about this! If I can take out the non-const calls, the class and test will be a lot more compact.

alexrlongne:
Looks good to me! I don't have any comments.

src/compton/Compton.hh
keadyk: I honestly forgot that you could call const member functions on non-const objects! I'm going to ditch all the non-const calls, since none of the functions are modifying state...

src/compton/Compton.cc
attom: For opac_type, how about 'imc' and 'sn'? Or 'monte_carlo' and 'deterministic'?

KineticTheory:
LGMT - I'm going to merge this now.  We will still have our discussion about the 4-D vector and the use of project names to toggle behavior.  Any changes that come out of that discussion can be put into a new PR.

src/compton/Compton.cc
keadyk: Those names are certainly on the drawing board-- my only concern is that they might be too generic. For instance, depending on implementation, an IMC code could use the opacities we call "S_N." That's why I defaulted to the code-specific naming convention :)

PR #237, opened 2017-06-15
A few small changes to the regression/CI scripts.

KineticTheory:
+ Add a knl build to the CI testing system.
+ Change how the number of processors available for running tests is set when using a Cray system.  We now use `lscpu` to look at the number of sockets and the number of cores per socket.
+ Set the slurm job name to be representative of project name and the pr number.
+ Clean up indentation of documentation in UnitTest.hh.
+ This PR captures code that is already _live_ at `/usr/projects/jayenne/regress`.
  * [ ] Travis CI checks pass

### DIFF
--- a/regression/draco_regression_macros.cmake
+++ b/regression/draco_regression_macros.cmake
@@ -31,6 +31,7 @@ macro( find_num_procs_avail_for_running_tests )
   if( DEFINED ENV{CRAYPE_DIR} )
 
     set( num_test_procs 1 )
+    set( num_sockets 1 )
     if( EXISTS /usr/bin/lscpu )
       execute_process( COMMAND /usr/bin/lscpu
         OUTPUT_VARIABLE lscpu_out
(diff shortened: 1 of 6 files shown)


regression/draco_regression_macros.cmake
```diff
@@ -31,26 +31,32 @@ macro( find_num_procs_avail_for_running_tests )
...
+    set( num_sockets 1 )
     if( EXISTS /usr/bin/lscpu )
       execute_process( COMMAND /usr/bin/lscpu
         OUTPUT_VARIABLE lscpu_out
         OUTPUT_STRIP_TRAILING_WHITESPACE )
       # break text block into a list of lines
       string( REPLACE "\n" ";"  lscpu ${lscpu_out} )
       foreach( line ${lscpu} )
+        if( ${line} MATCHES "Socket")
+          string( REGEX REPLACE "^.* ([0-9]+)$" "\\1" num_sockets ${line} )
```
alexrlongne: This section looks like it sets up regressions to run with multiple cores (both sockets and cores/socket are taken from lscpu output). I thought Trinitite could not run multiple tests concurrently on a node? Is this just the actual number of cores available for the case when tests need multiple cores?

KineticTheory: @alexrlongne Trinitite should allow multiple jobs to run concurrently on a node via the srun option `--gres=craynetwork:0 --exclusive`

PR #247, opened 2017-06-29
Fix plot2d tests that were failing on snow.

KineticTheory:
**Why are these changes needed:**

Snow has an older version of xmgrace than what is found on moonlight and the output format of the metadata section of the plot file wasn't matching the gold files. I reworked the logic in the tstPlot2D test to allow this test to pass on both moonlight and snow by omitting the whole metadata section of the plot file from the comparison check. Since the size of the meta data block varies between machines, I added logic to discover how many lines needed to be removed.

**What was changed:**

+ Updated the `FindGrace.cmake` script to deal with LANL HPC's directory structures:
  - `/usr/xmgrace`
  - `/lib64/libgrace_np.a`
  - `/var/lib/perceus/vnfs/compute/rootfs/usr/include/grace_np.h`
+ Update bench files by removing all of the metadata at the top of the file.
+ Update the `checkOutputFiles` function in `tstPlot2D.cc`
  - After the test generates a plot file, figure out how many lines are meta data and strip those lines from the file. The amount of data removed was previously hard coded (i.e.: the top 58 lines).
+ Refs: [#954](https://rtt.lanl.gov/redmine/issues/954)
  * [x] Travis CI checks pass


KineticTheory:
Trouble reading grace_np.h from the back-end.  I'm looking into it.

KineticTheory:
@along Can you review the build system changes?
@lowrie Can you review the changes to C++ code?

KineticTheory:
ugh.  Still broken on snow (seems to work from an interactive session).

config/FindGrace.cmake
```diff
@@ -6,77 +6,201 @@
...
+#  Grace_FOUND          - True if Grace found.
 #  Grace_INCLUDE_DIRS   - where to find Grace.h, etc.
 #  Grace_LIBRARIES      - List of libraries when using Grace.
-#  GRACE_FOUND          - True if Grace found.
+#  Grace_EXECUTABLE     - main program
+#
+# Hints
+# ^^^^^
+#
+# Set ``Grace_ROOT_DIR`` to a directory that contains a Grace installation.
```
lowrie: Maybe this is inconsistent with other draco environment settings, but grace uses GRACE_HOME for what I believe you're calling Grace_ROOT_DIR.  No big deal.

alexrlongne:
The build system part looks good. I guess if you don't use GRACE_HOME find_program will set Grace_ROOT_DIR correctly anyways.

config/FindGrace.cmake
KineticTheory: I can make this change.  I used `Grace_ROOT_DIR` because some other TPL used that naming convention.  It looks like Grace uses `GRACE_HOME` in their documentation, so I will switch it.

PR #245, opened 2017-06-28
WIP: Eliminate direct comparison of floating point types

KineticTheory:
Replaced by #251, #252, #253 and #254.

I built Draco with user provided `C_FLAGS='-Wfloat-equal -Werror' CXX_FLAGS='-Wfloat-equal -Werror'` compiler options (GCC) to discover all of the cases where Draco did a direct comparison of two floating point values.  I replaced such implementations with a safer comparison.

Also:
* When using GCC, add the following new flags `-Wunreachable-code -D_FORTIFY_SOURCE=2`
* Cleaned up comment blocks.
* Added const qualifiers in a few places.
* Use C++11 uniform initialization in a few places.

To Do:
* [ ] Test Jayenne and Capsaicin against this version of Draco before merge.

* [Pre-Merge Code Review](https://github.com/lanl/Draco/wiki/Style-Guide)
  * [ ] Travis CI checks pass
  * [ ] Code coverage does not decrease
  * [ ] [Valgrind test passes](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] [Toss2 checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] [Toss3 checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] [Trinitite checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] Code reviewed/approved, sufficient DbC checks, testing, documentation
    * [ ] Jae - `cdi, cdi_eospac, cdi_ipcress, mesh_element, meshReaders, quadrature, RTT_Format_Reader`
    * [ ] Kent - `fit, linear, memory, min, norms, ode, parser, roots, special_functions`
    * [ ] Alex - `c4, device, diagnostics, ds++, plot2D, rng, timestep, units, viz`
    * [ ] Kendra - `compton`

PR #251, opened 2017-07-14
Eliminate direct comparison of floating point types (c4, ds++, rng,...)

KineticTheory:
+ To improve code reproducibility between machines, I am attempting to eliminate direct comparisons of floating point types.
+ I built Draco with user provided C_FLAGS='-Wfloat-equal -Werror' CXX_FLAGS='-Wfloat-equal -Werror' compiler options (GCC) to discover all of the cases where Draco did a direct comparison of two floating point values. I replaced such implementations with a safer comparison.
+ Also in this changeset:
  - When using GCC, add the `-Wunreachable-code` flag.
  - Cleaned up comment blocks.
  - Added const qualifiers in a few places.
  - Use C++11 uniform initialization in a few places.
+ This is the same as #245, but broken into smaller chunks for review.
  * [x] Travis CI checks pass


KineticTheory:
@along This PR is now ready for final review.  All tests pass except:
* Travis/Codecov fails - but there are no substantial changes to actual code so I would like to ignore these failures.
* tstTime (3x) fails on trinitite.  Failures appear to be jitter.
```
Test: failed
     t.wall_clock() value does not match the expected value.
	end            = 1.50016e+09
	begin          = 1.50016e+09
	end-begin      = 2.57897
	t.wall_clock() = 3.94859
	error          = 1.36962
	prec           = 0.02
```

KineticTheory:
@alexrlongne PR#251 is ready for final review/merge.

src/ds++/test/tstField_Traits.cc
```diff
@@ -20,27 +21,27 @@ using namespace rtt_dsxx;
...
-    ut.passes("complex zero good");
+  if (rtt_dsxx::soft_equiv(value(x), 3.7))
+    PASSMSG("complex zero good");
   else
-    ut.failure("complex zero NOT good");
+    FAILMSG("complex zero NOT good");
 
-  if (Field_Traits<double const>::zero() == 0.0)
-    ut.passes("double zero good");
+  if (rtt_dsxx::soft_equiv(Field_Traits<double const>::zero(), 0.0))
```
alexrlongne: I think this zero could use the tightest tolerance available, maybe `numeric_limits<double>::min()` and the same thing for the check on line 42?

src/ds++/test/tstPacking_Utils.cc
```diff
@@ -305,10 +304,10 @@ void packing_test(rtt_dsxx::UnitTest &ut) {
     u.extract(2, cc);
 
     for (size_t i = 0; i < 50; ++i)
-      if (x[i] != 0)
+      if (!rtt_dsxx::soft_equiv(x[i], 0.0))
```
alexrlongne: Do you want to use `x[i] < numeric_limits<double>::min()` for comparisons to zero as is done elsewhere in this PR?

src/special_functions/KroneckerDelta.hh
```diff
@@ -5,23 +5,40 @@
...
+}
+
 template <typename T>
-DLL_PUBLIC_special_functions unsigned int kronecker_delta(T const test_value,
-                                                          T const offset);
+unsigned int kronecker_delta(
+    T const test_value, T const offset,
+    typename std::enable_if<std::is_floating_point<T>::value>::type * = 0) {
+  return rtt_dsxx::soft_equiv(test_value, offset, 1.0e-16) ? 1 : 0;
+}
```
KineticTheory: yes. fixed.

src/ds++/test/tstField_Traits.cc
KineticTheory: fixed.

src/ds++/test/tstPacking_Utils.cc
KineticTheory: yes.

PR #252, opened 2017-07-14
Eliminate direct comparison of floating point types (RTT_Format_Reade…

KineticTheory:
+ To improve code reproducibility between machines, I am attempting to eliminate direct comparisons of floating point types.
+ I built Draco with user provided C_FLAGS='-Wfloat-equal -Werror' CXX_FLAGS='-Wfloat-equal -Werror' compiler options (GCC) to discover all of the cases where Draco did a direct comparison of two floating point values. I replaced such implementations with a safer comparison.
+ Also in this changeset:
  - When using GCC, add the `-Wunreachable-code` flag.
  - Cleaned up comment blocks.
  - Added const qualifiers in a few places.
  - Use C++11 uniform initialization in a few places.
+ This is the same as #245, but broken into smaller chunks for review.
  * [ ] Travis CI checks pass

### DIFF
--- a/src/cdi/CDI.cc
+++ b/src/cdi/CDI.cc
@@ -10,6 +10,7 @@
 
 #include "CDI.hh"
 #include "ds++/Safe_Divide.hh"
+#include "ds++/Soft_Equivalence.hh"
 #include <iostream>
 #include <limits>
 #include <numeric>
@@ -267,7 +268,7 @@ void CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
 
   planck.resize(groups, 0.0);
 
-  if (T == 0)
+  if (rtt_dsxx::soft_equiv(T, 0.0, 1.0e-16))
     return;
 
   double scaled_frequency;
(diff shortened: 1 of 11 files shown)


src/cdi_analytic/Pseudo_Line_Base.cc
```diff
@@ -249,7 +249,8 @@ double Pseudo_Line_Base::monoOpacity(double const x, double const T) const {
       Result += edge_factor_[i] * cube(nu0 / x);
     }
   }
-  if (Tpow_ != 0.0) {
+  if (!rtt_dsxx::soft_equiv(Tpow_, 0.0,
+                            std::numeric_limits<double>::epsilon())) {
```
RyanWollaeger: Why is the `numeric_limits` function used here instead of 1.0e-16 (or vice versa)?

KineticTheory: The value reported by numeric_limits is more representative of the relative tolerance between doubles (I think for Intel Xeon it is 2.2e-16). `epsilon` is also based on the underlying hardware, so it will likely have a different value on Sierra, which uses IBM Power9 chips.

src/cdi/CDI.cc
```diff
@@ -267,7 +268,7 @@ void CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
 
   planck.resize(groups, 0.0);
 
-  if (T == 0)
+  if (rtt_dsxx::soft_equiv(T, 0.0, 1.0e-16))
```
RyanWollaeger: What about `numeric_limits` here and below?

KineticTheory:
@RyanWollaeger I made the changes you requested.  Can you merge this PR?

PR #256, opened 2017-07-18
Rework Global_Timer to be more useful in parallel runs.

kgbudge:
The existing implementation returned statistics only for processor 0 unless you went through gyrations that pretty much missed the whole reason for using Global_Timer rather than just Timer. The new implementation returns means and deviations for all non-PAPI statistics across processors.
    
    The new implementation also produces a warning message if timers you request to be active are never found during the run. The warning message tells you what timers were found that were not active, which can be very useful for figuring out what timer you really wanted, or where a new timer should be inserted in the code.

  * [ ] Travis CI checks pass

### DIFF
--- a/src/c4/Global_Timer.cc
+++ b/src/c4/Global_Timer.cc
@@ -88,14 +88,14 @@ void Global_Timer::reset_all() {
 /*static*/
 void Global_Timer::report_all(ostream &out) {
   if (rtt_c4::node() == 0) {
-    for (unsigned i = 0; i < 80; ++i)
+    for (unsigned i = 0; i < 92; ++i)
       cout << '-';
     cout << endl;
     cout << "Timing report for all global timers:" << endl << endl;
-    cout
-        << "N                      user                  system                "
-           "wall"
-        << endl;
+    cout << "         N                      user                   system     "
+            "           "
+            " wall"
+         << endl;
   }
   bool deferred_not_found = false;
   for (auto const &i : active_list_) {
@@ -130,7 +130,7 @@ void Global_Timer::report_all(ostream &out) {
         }
       }
     }
-    for (unsigned i = 0; i < 80; ++i)
+    for (unsigned i = 0; i < 92; ++i)
       cout << '-';
     cout << endl;
   }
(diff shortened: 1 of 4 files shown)


KineticTheory:
@kgbudge I also want Alex to review this since he is using some of the timer tools from Draco.

src/c4/Timer.cc
```diff
@@ -321,6 +322,57 @@ void Timer::pause(double const pauseSeconds) {
...
+  double s = sum_system_cpu(), s2 = s * s;
+  rtt_c4::global_sum(s);
+  rtt_c4::global_sum(s2);
+  double ms = s / ranks;
+
+  double ww = sum_wall_clock(), ww2 = ww * ww;
+  rtt_c4::global_sum(ww);
+  rtt_c4::global_sum(ww2);
+  double mww = ww / ranks;
+
```
KineticTheory: Would there be any advantage to packing `{ni, ni2, u, u2, s, s2, ww, ww2}` into an array/vector to limit the function to a single `global_sum`?

src/c4/Timer.cc
```diff
@@ -321,6 +322,57 @@ void Timer::pause(double const pauseSeconds) {
...
+        << setw(w) << mu << " +/- "
+        << sqrt((s2 - 2 * ms * s + ranks * ms * ms) / ranks) << setw(w) << mu
+        << " +/- " << sqrt((ww2 - 2 * mww * ww + ranks * mww * mww) / ranks);
+
+    // Omit PAPI for now.
+
+    out << std::endl;
+
+    out.flush();
+  }
```
KineticTheory: * I thought `std::endl` always flushed the buffer (but maybe I'm wrong).  If so, then this call to `flush()` shouldn't be required.

src/c4/Global_Timer.cc
```diff
@@ -130,7 +130,7 @@ void Global_Timer::report_all(ostream &out) {
         }
       }
     }
-    for (unsigned i = 0; i < 80; ++i)
+    for (unsigned i = 0; i < 92; ++i)
       cout << '-';
```
KineticTheory: New code was added to this PR, so I will review the new portions:
* Maybe use `cout << string(92,'-'); ` instead of the 2 loops above.

PR #261, opened 2017-07-24
Fix a bug introduced in #252 (early return if T==0).

KineticTheory:
I inadvertantly introduced a bug into CDI when I replaced:
```
if (T == 0.0 )
```
with
```
if (soft_equiv(T,0.0,numeric_limits<double>::epsilon())
```
This check is needed to avoid a division by zero.  However, in CDI `T` is almost always < epsilon() because of the unit system. Thus, the Planck and Rosseland integrator functions were returning earlier than they should have.  The fix in this PR is to use the conditional
```
if ( T < max(nu) * numeric_limits<double>::min() )
```
which should guarantee that `nu/T < numeric_limits<double>::max()`.  Also add a test for very small values of T.

* Fixes #252
  * [ ] Travis CI checks pass

### DIFF
--- a/src/cdi/CDI_integratePlanckSpectrum.cc
+++ b/src/cdi/CDI_integratePlanckSpectrum.cc
@@ -21,9 +21,9 @@ namespace rtt_cdi {
  * and high must first be multiplied by Planck's constant and temperature by
  * Boltzmann's constant before they are passed to this function.
  *
- * \param low lower frequency bound.
- * \param high higher frequency bound.
- * \param T the temperature (must be greater than 0.0)
+ * \param[in] low lower frequency bound.
+ * \param[in] high higher frequency bound.
+ * \param[in] T the temperature (must be greater than 0.0)
  *
  * \return integrated normalized Plankian from low to high
  */
@@ -32,8 +32,9 @@ double CDI::integratePlanckSpectrum(double low, double high, const double T) {
   Require(high >= low);
   Require(T >= 0.0);
 
-  // return 0 if temperature is a hard zero
-  if (rtt_dsxx::soft_equiv(T, 0.0, std::numeric_limits<double>::epsilon()))
+  // high/T must be < numeric_limits<double>::max().  So, if T ~< high*min, then
+  // return early with zero values (assuming max() ~ 1/min()).
+  if (T < high * std::numeric_limits<double>::min())
     return 0.0;
 
   // Sale the frequencies by temperature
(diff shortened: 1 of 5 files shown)


src/cdi/test/tCDI.cc
```diff
@@ -678,6 +678,23 @@ void test_planck_integration(rtt_dsxx::UnitTest &ut) {
...
+        CDI::integratePlanckSpectrum(0.1, 1.0, 1.0e-30);
+    if (!soft_equiv(integrate_range_cold, 0.0))
+      ITFAILS;
+  }
+  // Try an extreme case 2. T < numeric_limits<double>::min() --> follow special
+  // logic and return zero.
+  {
+    double const integrate_range_cold =
+        CDI::integratePlanckSpectrum(0.1, 1.0, 1.0e-308);
+    if (!soft_equiv(integrate_range_cold, 0.0))
```
RyanWollaeger: Could this check be made more restrictive, since `integratePlanckSpectrum` should return 0.0?
Would something like `rtt_dsxx::soft_equiv(integrate_range_cold, 0.0, std::numeric_limits<double>::epsilon())` work?

src/cdi/CDI_integratePlanckSpectrum.cc
```diff
@@ -32,8 +32,9 @@ double CDI::integratePlanckSpectrum(double low, double high, const double T) {
   Require(high >= low);
   Require(T >= 0.0);
 
-  // return 0 if temperature is a hard zero
-  if (rtt_dsxx::soft_equiv(T, 0.0, std::numeric_limits<double>::epsilon()))
+  // high/T must be < numeric_limits<double>::max().  So, if T ~< high*min, then
+  // return early with zero values (assuming max() ~ 1/min()).
+  if (T < high * std::numeric_limits<double>::min())
```
RyanWollaeger: This is a very nice way to do the check!
Is there any motivation to use `std::numeric_limits<double>::epsilon()`, since it has now been non-dimensionalized?

KineticTheory: I don't think I can use `epsilon()` here.  My thinking was to change the limiting factor from "don't divide by zero" to "ensure division by T is representable."

lowrie: Later, effectively exp(-high/T) is called when integrate_planck() is called.  So then it's roughly exp(-1.0/min), which I'm guessing it underflows.  This probably gets set to zero, depending on the compiler options.

KineticTheory: Would you recommend that I change the 'early return' option conditional to 
```
if (T < - (low/ln(numeric_limits<double>::min()))) {...}
```

That expression is derived from requiring `exp(-nu/T) > min()`

PR #285, opened 2017-08-28
Provide common string functions in a single file.

KineticTheory:
Jayenne is refactoring some of Milagro.  To assist with this update, several new string manipulation features have been added to ds++.

**New functions:**

`trim` - remove whitespace (or any specified characters) from beginning and end of a string.
`prune` - remove specified characters from a string.
`parse_number` - Convert a string into a floating-point or integral type (int, long, unsigned, float, double, etc.)
`string_to_numvec` - Convert a string into a vector<T> where T is a numeric type. String entries are separated by specified delimiter. Optional beginning and ending range symbols may also be provided.  For example, `string = "{ 1.3, 2.3 }" --> vector<float> = {1,3, 2.3};`

**Relocated functions:**

`to_string` - convert a number into a string preserving precision. Previously found at `ds++/to_string.hh`
`tokenize`  - convert a string into a vector<string> based on the specified delimiter. Previously found at `UnitTest.hh.`
`get_word_count` - convert a `stringstream` or `fstream` into a `map<string,unsigned>` where the number of occurrences of each word is reported. Previously found at `UnitTest.hh`. Often used to check text output from unit tests.

**Other Changes:**

+ `to_string*` classes and tests removed (absorbed by `DracoStrings*`).
+ Update include directives to load `DracoStrings.hh` instead of `to_string.hh`.
+ Minor code modernization in Viz.
+ Cleanup comment blocks, copyright dates, etc.

  * [x] Travis CI checks pass


keadyk:
Looking this over now!

src/ds++/DracoStrings.hh
```diff
@@ -0,0 +1,166 @@
...
+ * \param[in] num The integral or float-point type number that will be converted
+ *                to a string.
+ * \param[in] precision How many digits will be preserved?
+ * \return A string representation of the numeric value
+ *
+ * \sa http://public.research.att.com/~bs/bs_faq2.html
+ */
+template <typename T>
+std::string to_string(T const num, unsigned int const precision = 23) {
+  std::stringstream s;
```
keadyk: I suppose you could be pedantic and use ostringstream here?

src/ds++/DracoStrings.hh
```diff
@@ -0,0 +1,166 @@
...
+/*!
+ * \brief trim whitespace (or other characters) from before and after main text.
+ *
+ * \param[in] str The string that will be processed
+ * \param[in] whitespace A set of characters that will be removed.
+ * \return A new, probably shortened, string without unwanted leading/training
+ *         characters.
+ */
+DLL_PUBLIC_dsxx std::string trim(std::string const &str,
+                                 std::string const &whitespace = " \t");
```
keadyk: I don't see anything specific in our style guide about this, but perhaps it would be useful to describe default values in the doxygen comments?

src/ds++/DracoStrings.cc
```diff
@@ -0,0 +1,175 @@
...
+ *        each.
+ */
+std::map<std::string, unsigned> get_word_count(std::ostringstream const &msg,
+                                               bool verbose) {
+  using std::map;
+  using std::string;
+  using std::cout;
+  using std::endl;
+
+  map<string, unsigned> word_list;
```
keadyk: (I didn't realize maps auto-initialized data when a key was added, so I learned something new today!)

src/ds++/DracoStrings.cc
```diff
@@ -0,0 +1,175 @@
...
+#include <fstream>
+#include <iostream>
+
+namespace rtt_dsxx {
+
+//----------------------------------------------------------------------------//
+// Definitions for fully specialized template functions
+//----------------------------------------------------------------------------//
+
+template <> auto parse_number<int>(std::string const &str) -> int {
```
keadyk: Do we want any sort of error- or input-checking here? For instance, I think stoi could read something like "123abc" and return 123-- is this a valid use case?

src/ds++/DracoStrings.hh
KineticTheory: I updated https://rtt.lanl.gov/redmine/projects/draco/wiki/Coding_Standards#section-23 with an example.

PR #292, opened 2017-09-07
Upgrade how the Compton TPL is discovered, presented and used.

KineticTheory:
This change set upgrades CSK to a full Draco TPL and publishes it via the installed `draco-config.cmake` file.

+ Add the compton package to the `.clang-format` list of Draco packages (for sorting of header files).
+ Complete rewrite of `FindCOMPTON.cmake` makes use of target properties instead of environment and cmake variables.
+ Move the discovery of compton from `src/compton/CMakeLists.txt` to `config/vendor_libraries.cmake`
+ Always install a *compton* package when building Draco.  If the CSK library is not found, the installation will be a stubbed out `Compton.hh` file and no library.
+ A `compton/config.h` file has been added to capture the CPP macro `COMPTON_FOUND`. Include directives for CSK files were moved to this new file. This mimics the way we treat other TPLs like Random123 or Metis.
+ Updated the Lib_compton source files to make proper use of the above changes.
+ This will break Jayenne until it is updated to use this new TPL API.

  * [ ] Travis CI checks pass (failing due to #293)
  * [ ] [Trinitite checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco) (not tested due to 2-week outage)

### DIFF
--- a/config/FindCOMPTON.cmake
+++ b/config/FindCOMPTON.cmake
@@ -6,86 +6,192 @@
 # note   Copyright (C) 2016-2017 Los Alamos National Security, LLC.
 #        All rights reserved.
 #------------------------------------------------------------------------------#
-# - Find Compton
-# Find the native Compton includes and library
-#
-#  COMPTON_INCLUDE_DIRS   - where to find multigroup_lib_builder.hh, etc.
-#  COMPTON_LIBRARIES      - List of libraries when using Compton.
-#  COMPTON_FOUND          - True if Compton found.
 
-if (APPLE)
-    set(CMAKE_FIND_LIBRARY_SUFFIXES ".a;.dylib")
+#.rst:
+# FindCOMPTON
+# ---------
+#
+# Find the CSK COMPTON includes and libraries.
+#
+# CSK COMPTON is a project dedicated to the calculation and interpolation of
+# Compton Scattering Kernel (CSK, hence the project name) values for use in
+# radiative transfer simulations.  https://gitlab.lanl.gov/keadyk/CSK_generator
+#
+# Imported Targets
+# ^^^^^^^^^^^^^^^^
+#
+# If COMPTON is found, this module defines the following :prop_tgt:`IMPORTED`
+# targets::
+#
+#  COMPTON::compton         - The COMPTON library.
+#
+# Result Variables
+# ^^^^^^^^^^^^^^^^
+#
+# This module will set the following variables in your project::
+#
+#  COMPTON_FOUND          - True if COMPTON found on the local system
+#  COMPTON_INCLUDE_DIRS   - Location of COMPTON header files.
+#  COMPTON_LIBRARIES      - The COMPTON libraries.
+#  COMPTON_VERSION        - The version of the discovered COMPTON install.
+#
+# Hints
+# ^^^^^
+#
+# Set ``COMPTON_ROOT_DIR`` to a directory that contains a COMPTON installation.
+#
+# This script expects to find libraries at ``$COMPTON_ROOT_DIR/lib`` and the
+# COMPTON headers at ``$COMPTON_ROOT_DIR/include``.  The library directory may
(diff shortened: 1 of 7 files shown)


src/compton/CMakeLists.txt
```diff
@@ -9,67 +9,70 @@
...
-include_directories(
-   ${PROJECT_SOURCE_DIR}      # component headers
-   ${PROJECT_BINARY_DIR}      # compton/config.h
-   ${PROJECT_SOURCE_DIR}/..   # ds++ and compton header files
-   ${dsxx_BINARY_DIR}         # ds++/config.h
-   ${ode_BINARY_DIR}          # ode/config.h
-   ${COMPTON_INCLUDE_DIR}
+  include_directories(
+    ${Draco_SOURCE_DIR}/src    # compton/Compton.hh ds++/Assert.hh
+    ${Draco_BINARY_DIR}/src    # compton/config.h
```
keadyk: Is there a reason you prefer this syntax to ${PROJECT_BINARY_DIR}, which would let you include the config.h file without the enclosing compton/ folder? Is it just for clarity, so the #includes in Compton.* are clearly spelled out as "compton/<file>", or is there another reason?

KineticTheory: * The `include_directories` is deprecated in modern CMake.  I will slowly reduce/eliminate these calls.
* I have started using `${Draco_SOURCE_DIR}/src` to reduce the number of `-I<dir>` commands that appear on the compile line. This change also normalizes directory layout between the build directory and the install directory.

config/vendor_libraries.cmake
```diff
@@ -588,20 +588,68 @@ macro( setupSuperLU_DIST )
...
+      PURPOSE "SuperLU is a general purpose library for the direct solution of
+large, sparse, nonsymmetric systems of linear equations on high performance
+machines."  )
+
+  endif()
+
+endmacro()
+
+#------------------------------------------------------------------------------
+# Setup COMPTON (https://gitlab.lanl.gov/csk)
```
keadyk: I think the url should be gitlab.lanl.gov/keadyk/CSK_generator

config/FindCOMPTON.cmake
```diff
@@ -6,86 +6,192 @@
...
+
+# Include these modules to handle the QUIETLY and REQUIRED arguments.
+include(FindPackageHandleStandardArgs)
 
-if (APPLE)
-    set(CMAKE_FIND_LIBRARY_SUFFIXES ".a;.dylib")
+#=============================================================================
+# If the user has provided ``COMPTON_ROOT_DIR``, use it!  Choose items found
+# at this location over system locations.
+if( EXISTS "$ENV{COMPTON_ROOT_DIR}" )
```
keadyk: This variable should be set by the module file, right? Is that what you mean by "if the user has provided COMPTON_ROOT_DIR"? (I just want to be sure I understand what's going on.)

keadyk:
The CI build failed, but it appears to be the system's fault, not the PR.

config/FindCOMPTON.cmake
KineticTheory: The module sets this variable (or it could be set w/o a module).  Ideally, the module should set `CMAKE_PREFIX_PATH` instead.  A CMake variable of this same name will also work.

KineticTheory:
Travis checks are failing due to a security issue in old python files. I'll open a different ticket to fix this.

keadyk:
Alright! Would you like me to wait to merge until the Travis issues are fixed, or can I proceed now?

KineticTheory:
@keadyk Please review https://gitlab.lanl.gov/jayenne/jayenne/merge_requests/133 first.  If that looks okay, then merge #292 w/o requiring travis to pass.

keadyk:
Roger that!

keadyk:
(Sorry, wrong button)

PR #280, opened 2017-08-22
Expand MPI functionality to simplify logic in Jayenne

alexrlongne:
* Add more MPI functions to Draco's C4 library and allow custom types in send/receives
  * Add non-blocking sends and receives with custom MPI datatypes
  * Add no-blocking allreduce call with the standard of
  * send_is test now includes example of a custom MPI datatype and a send/receive with that type
  * tstReduction now includes non-blocking allreduce with several types

* Purpose of Pull Request
  * Allow MPI messaging for any class without packing and unpacking into character streams
  * Allow Jayenne to use the simpler non-blocking allreduce instead of complicated reduction

* [Pre-Merge Code Review](https://github.com/lanl/Draco/wiki/Style-Guide)
  * [ ] Travis CI checks pass
  * [ ] Code coverage does not decrease
  * [ ] [Valgrind test passes](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] [Toss2 checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] [Toss3 checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] [Trinitite checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] Code reviewed/approved, sufficient DbC checks, testing, documentation
  * [ ] Ensure `cmake -DDRACO_C4=SCALAR ~/draco` works (It doesn't.  Not even before this PR.  KT will help with fixing this issue.)

PR #315, opened 2017-10-03
Allow c4 to build when DRACO_C4=SCALAR.

KineticTheory:
+ While reviewing #280, I discovered that Draco no-MPI builds were broken.  This PR fixes `develop` for no-MPI builds and should be merged before #280.
+ Update function signature for `T prefix_sum` found in `C4_Serial.hh` to use the same argument types as the function of the same name found in `C4_MPI.hh`. This fixes the build error `call of overloaded function is ambiguous`.
+ Update `tstScalar.cc` to check the value of `request.complete()` instead of `request.count()` since the later member function no longer exists.
+ Clean up comments.
  * [ ] Travis CI checks pass


alexrlongne:
I'm going to merge this. I think the decrease in test coverage is because it can't see the serial path.

PR #332, opened 2017-10-18
Add [[noreturn]] attribute to non-returning Assert.hh functions.

kgbudge:
Added the [[noreturn]] attribute to insist, insist_ptr, toss_cookies, and toss_cookies_ptr.

This potentially eliminates spurious warnings. See tstAssert.cc for an example of a function where a spurious "return with no value" warning is eliminated.

  * [x] Travis CI checks pass
  * [x] [Valgrind test passes](https://rtt.lanl.gov/CDash/index.php?project=Draco) - 2 valgrind issues reported - but not related to this changeset.

### DIFF
--- a/src/ds++/Assert.hh
+++ b/src/ds++/Assert.hh
@@ -169,12 +169,12 @@ public:
 //---------------------------------------------------------------------------//
 
 //! Throw a rtt_dsxx::assertion for Require, Check, Ensure.
-DLL_PUBLIC_dsxx[[noreturn]] void
+[[noreturn]] DLL_PUBLIC_dsxx void
 toss_cookies(std::string const &cond, std::string const &file, int const line);
 
-DLL_PUBLIC_dsxx[[noreturn]] void toss_cookies_ptr(char const *const cond,
-                                                  char const *const file,
-                                                  int const line);
+[[noreturn]] DLL_PUBLIC_dsxx void toss_cookies_ptr(char const *const cond,
+                                                   char const *const file,
+                                                   int const line);
 
 //! Throw a rtt_dsxx::assertion if condition fails
 DLL_PUBLIC_dsxx void check_cookies(bool cond, char const *cond_text,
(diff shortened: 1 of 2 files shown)


KineticTheory:
@kgbudge I'm not familiar with this attribute.  I want to read about it before I merge.  I also want to make sure your implementation works correctly with MSVC.

src/ds++/Assert.hh
```diff
@@ -169,11 +169,12 @@ public:
 //---------------------------------------------------------------------------//
 
 //! Throw a rtt_dsxx::assertion for Require, Check, Ensure.
-DLL_PUBLIC_dsxx void toss_cookies(std::string const &cond,
-                                  std::string const &file, int const line);
+DLL_PUBLIC_dsxx[[noreturn]] void
```
KineticTheory: I want to read about this syntax before I merge your PR.

kgbudge: I think you should. This is new ground being broken.

C++11 standardizes attributes, which are (usually platform-dependent) annotations to conventional C++ constructs. They seem to be an attempt to get away from pragmas. As with pragmas, the grammar for attributes is very flexible and attributes a platform does not recognize are supposed to be ignored.

A small number of attributes are standardized and so are supposed to be recognized by all platforms. [[noreturn]] on a function declaration is one of the standard attributes, and its meaning is that the function will never return (because it invariably either terminates the program or throws an exception.) A number of standard C++11 library functions, such as exit() and abort(), are required to have this attribute. Our toss_cookies and insist functions are in the same category since they always throw an exception. 

The idea of [[noreturn]] is that the flow analyzer of a conforming C++11 compiler will know that a call to a [[noreturn]] function ends flow. So if a flow path in a function returning a value enters a [[noreturn]] function, and there is no return statement afterwards, the compiler knows that no "return without value" warning is required. An example where this is actually an issue is provided by the "unused" function I added to tstAssertion, which has an Insist(false, "bad case") as the default branch in a case statement. This seemingly produces a flow path with a return without value, which normally causes a warning, and so in the past I've always had to put a spurious return statement that can never be reached after this call. With the [[noreturn]] attribute, this warning is suppressed and no such call is needed.

I've confirmed that, in a case where a [[noreturn]] statement *is* followed by a spurious return, the compiler does *not* warn about that. But I've only tested it on our local platforms -- I am very curious whether we get consistent behavior across all our platforms. If not, this attribute becomes a lot less useful, and I'm willing to rethink this merge.

KineticTheory:
@kgbudge I'm good with this change.  However, the syntax isn't quite right for Visual Studio.  Please change the declaration to:

```cpp
[[noreturn]] DLL_PUBLIC_dsxx void [...]
```

* The `[[noreturn]]` must precede `DLL_PUBLIC_dsxx`
* There must be a space between the two *declspec* directives.

PR #378, opened 2018-03-28
Add a runtime_check function to rtt_c4

kgbudge:
Add a runtime_check function to rtt_c4 for synchronous check of local… condition.

This function should be called by all processors synchronously. It checks a condition which, if it is not satisfied on any processor, causes all processors to throw std::runtime_error with an appropriate message.

### Background

It is common for a parallel code to check some local condition that must be satisfied on all processors. If any processor fails, then all should throw an exception to be handled at some higher level of the code. If the program throws the exception only on those processors that fail, recovery is difficult and may leave the user relying on MPI to notice that a process has died, or, worse yet, MPI hangs waiting for a message that will never arrive.

### Purpose of Pull Request

Add  runtime_error.hh, runtime_errror.cc to the c4 package to define a synchronous runtime check function. Add test/tstruntime_error to exercise this function.

* Reference:
* Checks
  * [ ] Travis CI checks pass

### DIFF
--- a/src/c4/runtime_check.cc
+++ b/src/c4/runtime_check.cc
@@ -1,0 +1,36 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   ds++/runtime_check.cc
+ * \author Kent Grimmett Budge
+ * \date   Wed Mar 28 07:58:48 2018
+ * \brief  Member definitions of class runtime_check
+ * \note   Copyright (C) 2018 Los Alamos National Security, LLC.
+ *         All rights reserved. */
+//---------------------------------------------------------------------------//
+
+#include <exception>
+
+#include "C4_Functions.hh"
+#include "ds++/DracoStrings.hh"
+
+namespace rtt_dsxx {
+
+//-------------------------------------------------------------------------//
+void runtime_check(bool condition, char const *message) noexcept(false) {
+  Require(message != nullptr);
+
+  unsigned sum = !condition;
+  rtt_c4::global_sum(sum);
+  if (sum != 0) // some processors failed the condition
+  {
+    throw std::runtime_error("runtime error detected on " +
+                             rtt_dsxx::to_string(sum) + " processor(s): " +
+                             message);
+  }
+}
+
+} // end namespace rtt_dsxx
+
+//---------------------------------------------------------------------------//
+// end of ds++/runtime_check.cc
+//---------------------------------------------------------------------------//
--- a/src/c4/runtime_check.hh
+++ b/src/c4/runtime_check.hh
@@ -1,0 +1,47 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   ds++/runtime_check.hh
+ * \author Kent Grimmett Budge
+ * \brief  Define runtime_check function
+ * \note   Copyright (C) 2018 Los Alamos National Security, LLC.
+ *         All rights reserved. */
+//---------------------------------------------------------------------------//
+
+// clang-format off
+
+#ifndef dsxx_runtime_check_hh
+#define dsxx_runtime_check_hh
+
+namespace rtt_dsxx {
+
+//---------------------------------------------------------------------------//
+  /*! Parallel synchronous runtime check
+   *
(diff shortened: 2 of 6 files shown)


src/c4/runtime_check.cc
```diff
@@ -0,0 +1,36 @@
...
+#include <exception>
+
+#include "C4_Functions.hh"
+#include "ds++/DracoStrings.hh"
+
+namespace rtt_dsxx {
+
+//-------------------------------------------------------------------------//
+void runtime_check(bool condition, char const *message) noexcept(false) {
+  Require(message != nullptr);
```
KineticTheory: Does it make sense to make this function header-only and inline?

src/c4/runtime_check.cc
```diff
@@ -0,0 +1,36 @@
...
+ *         All rights reserved. */
+//---------------------------------------------------------------------------//
+
+#include <exception>
+
+#include "C4_Functions.hh"
+#include "ds++/DracoStrings.hh"
+
+namespace rtt_dsxx {
+
```
KineticTheory: Since this function doesn't wrap a MPI functionality, I think it should live in the `src/diagnostics` directory instead of `src/c4`.  
(The same can be said of other files/functions in c4, like ParallelUnitTest.  I'll add a personal task to review those files and move them.)

src/c4/runtime_check.cc
```diff
@@ -0,0 +1,36 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   ds++/runtime_check.cc
+ * \author Kent Grimmett Budge
```
KineticTheory: When you move these files to `diagnostics` please globally replace "ds++" with "diagnostics". Ditto for "dsxx" (namespace, etc.).

src/c4/runtime_check.cc
```diff
@@ -0,0 +1,36 @@
...
+
+  unsigned sum = !condition;
+  rtt_c4::global_sum(sum);
+  if (sum != 0) // some processors failed the condition
+  {
+    throw std::runtime_error("runtime error detected on " +
+                             rtt_dsxx::to_string(sum) + " processor(s): " +
+                             message);
+  }
+}
```
KineticTheory: Is there any value it making this look more like a Design-by-Contract assertion?  It would be easy to add CPP macros that wrap this function.  Something like `Parallel_Require( a>0 )` or `pInsist( b>0, "b must be positive )`.  The `global_sum` could be costly and might need to be turned off for fully optimized builds.  If you go this route, consider renaming these files to be variants of `Assert.hh` -- something like `Parallel_Assert.hh` or `MPI_Assert.hh`.

src/c4/runtime_check.cc
kgbudge: There's enough going on that I hesitate. Since this function is not suitable for calling within loops or other compute-intensive code (because it must be synchronous) it's not compute critical. The construction of the throw message takes some code.

src/c4/runtime_check.cc
kgbudge: I think that's a reasonable suggestion.

src/c4/runtime_check.cc
kgbudge: I did not originally write this with DBC in mind. I was thinking of actual run time errors, such as a file opening or write that fails, or bad user input, or a calculation that is well posed that runs off the rails (temperatures go negative or a matrix becomes numerically singular.) We want these things in the production builds, not just for debugging.  However, it's not unreasonable to think about a DBC wrapper for synchronous tests that we don't necessarily want turned on in production code.

KineticTheory: I think your UseCase is reasonable.  The function could be called directly (as you intended) or via a `mpiInsist` CPP macro (these are never disabled) but has the advantage of having a familiar name.

src/c4/runtime_check.cc
KineticTheory: I'll leave that decision up to you.  Anyway, it is doubtful that the function would ever be inlined because of the embedded `throw`.

PR #412, opened 2018-05-31
Add a parallel output stream.

kgbudge:
Define a class opstream and a global object of this class, pout, corresopnding to cout. This class behaves like any other output stream in the C++ stream library, with full support for formatting of output, manipulators, unformatted output, &c. However, output inserted by any MPI rank into the stream is buffered and does not appear on the console (cout) until the send() function is called on pout. At this point, all the buffered output for rank 0 since the last call to send() appears at the console, followed by all the output for rank 1, and so on. There is absolutely no requirement that the output from any rank be similar to that from any other rank, or that a rank produce any output at all. However, the send() function *must* be called synchronously; it is effectively a global_barrier.

This class is meant to have some of the functionality of the old Spin_Lock class, which was designed to serialize sections of code. The most common use case turned out to be serializing output, which Spin_Lock *did not actually achieve*. For this reason, Spin_Lock was abandoned.

The new opstream class is not useful except for serializing output, but it actually succeeds in this, and this is a very important use case for code serialization.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass


KineticTheory:
@kgbudge @alexrlongne This PR needs a little more attention.  There are a [handful of tests that fail](https://rtt.lanl.gov/cdash/index.php?project=Draco&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=-pr). Ignore the build warnings for the Coverage build. That one is on me.  I also have a handful of small changes that would like to see in the sources (mostly related to namespace injection and formatting). I'll provide those comments via the review process later today.

kgbudge:
K.T.,

Can you provide the comments on the changes you want? I would really like this capability for some diagnostic coding I'd like to merge with my project.

alexrlongne:
It looks like the C++ standards committee is also trying to address this issue, but we don't have C++ 20 support yet so I think we should use your solution.

https://en.cppreference.com/w/cpp/io/basic_syncbuf

kgbudge:
It would hardly be the first time we have anticipated an upcoming C++ solution.

I will look at the C++20 spec and see if I can tweak this one to conform a bit more, but my superficial impression is that they're already close.

kgbudge:
Okay, I looked the C++20 proposal over, and it's not really the same concept. The C++20 proposal is aimed at multithreading and does not offer anything for MPI synchronization.

src/c4/ofpstream.cc
```diff
@@ -0,0 +1,70 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   c4/ofpstream.cc
+ * \author Mike Buksas
+ * \date   Thu May  1 14:42:10 2008
+ * \brief
+ * \note   Copyright (C) 2016-2018 Los Alamos National Security, LLC.
+ *         All rights reserved.
```
KineticTheory: * Please update all of this header stuff (author, creation date, etc.)
* Ditto for the .hh file.

src/c4/ofpstream.hh
```diff
@@ -0,0 +1,86 @@
...
+public:
+  ofpstream(std::string const &filename);
+
+  void send() { sb_.send(); }
+  void shrink_to_fit() { sb_.shrink_to_fit(); }
+
+private:
+  struct mpibuf : public std::streambuf {
+
+    ~mpibuf() { send(); }
```
KineticTheory: Can `send()` throw? (that would be bad). Can we protect against this?

src/c4/ofpstream.hh
```diff
@@ -0,0 +1,86 @@
...
+ * blocks of output being generated, then remain there through the rest
+ * of the scope containing the stream, unless shrink_to_fit is called.
+ *
+ * \example c4/test/tstofpstream.cc
+ */
+//===========================================================================//
+
+class ofpstream : public std::ostream {
+public:
+  ofpstream(std::string const &filename);
```
KineticTheory: * Prefix with `explicit` to prevent addition of automatic promotion path from `std::string` to `ofpstream`.
* Prefix with `DLL_PUBLIC_c4`.

src/c4/ofpstream.hh
```diff
@@ -0,0 +1,86 @@
...
+
+  void send() { sb_.send(); }
+  void shrink_to_fit() { sb_.shrink_to_fit(); }
+
+private:
+  struct mpibuf : public std::streambuf {
+
+    ~mpibuf() { send(); }
+
+    void send();
```
KineticTheory: Please add some doxygen-based documentation for each member function.  This can be done in the .hh or the .cc file.

src/c4/ofpstream.hh
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * Output stream for serializing output from all MPI ranks.
+ *
+ * This class is an output stream that serialized output from all MPI ranks.
+ * Output is inserted to the stream as with any other output stream. When the
+ * send() function is invoked, which *must be done synchronously across all
+ * processors*, all output inserted to the stream since the last call to send()
+ * is printed to the associated file, by MPI rank. In other words, all output
+ * inserted to an instance of ofpstream by rank 0 since the last call to send()
+ * is printed to the associated file, then all output inserted to the instance of
```
KineticTheory: This line is > 80 cols. Please fix. (Emacs: move mouse to this paragraph; M-q)

src/c4/ofpstream.hh
```diff
@@ -0,0 +1,86 @@
...
+ * \note   Copyright (C) 2016-2018 Los Alamos National Security, LLC.
+ *         All rights reserved.
+ */
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+
+#ifndef c4_ofpstream_hh
+#define c4_ofpstream_hh
+
```
KineticTheory: Add `#include "c4/config.h"` so that `DLL_PUBLIC_c4` is defined (see comments below).

src/c4/opstream.cc
```diff
@@ -0,0 +1,61 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   c4/opstream.cc
+ * \author Mike Buksas
+ * \date   Thu May  1 14:42:10 2008
+ * \brief
+ * \note   Copyright (C) 2016-2018 Los Alamos National Security, LLC.
+ *         All rights reserved.
+ */
```
KineticTheory: Please update info in the header block.

src/c4/opstream.hh
```diff
@@ -0,0 +1,86 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   c4/opstream.hh
+ * \author Mike Buksas
+ * \brief  Define class opstream
+ * \note   Copyright (C) 2016-2018 Los Alamos National Security, LLC.
+ *         All rights reserved.
+ */
+//---------------------------------------------------------------------------//
```
KineticTheory: Please update information in the header block

src/c4/opstream.hh
```diff
@@ -0,0 +1,86 @@
...
+ * A stream of this type can be created only after MPI is initialized, and it
+ * must be  destroyed before MPI is shut down.
+ *
+ * \example c4/test/tstopstream.cc
+ */
+//===========================================================================//
+
+class opstream : public std::ostream {
+public:
+  opstream() : std::ostream(&sb_) {}
```
KineticTheory: Please provide doxygen-style comments for each member function.  This can be done in either the `.hh` or the `.cc` file.

src/c4/test/tstofpstream.cc
```diff
@@ -0,0 +1,59 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   c4/test/tstofpstream.cc
+ * \author Kent Budge
+ * \date   Wed Apr 28 09:31:51 2010
+ * \brief  Test c4::determinate_swap and c4::indeterminate_swap functions
+ * \note   Copyright (C) 2016-2018 Los Alamos National Security, LLC.
+ *         All rights reserved.
```
KineticTheory: Please update the comment block to match the content of the file.

src/c4/opstream.hh
```diff
@@ -0,0 +1,86 @@
...
+ *         All rights reserved.
+ */
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+
+#ifndef c4_opstream_hh
+#define c4_opstream_hh
+
+#include <vector>
```
KineticTheory: Add `#include "c4/config.h" so that `DLL_PUBLIC_c4` is defined.

src/c4/opstream.hh
```diff
@@ -0,0 +1,86 @@
...
+
+  void send() { sb_.send(); }
+  void shrink_to_fit() { sb_.shrink_to_fit(); }
+
+private:
+  struct mpibuf : public std::streambuf {
+
+    ~mpibuf() { send(); }
+
+    void send();
```
KineticTheory: Prefix with `DLL_PUBLIC_c4`.

src/c4/opstream.hh
```diff
@@ -0,0 +1,86 @@
...
+    void shrink_to_fit();
```
KineticTheory: Prefix with `DLL_PUBLIC_c4`.

src/c4/opstream.hh
```diff
@@ -0,0 +1,86 @@
...
+
+    virtual int_type overflow(int_type c);
```
KineticTheory: Prefix with `DLL_PUBLIC_c4`.

src/c4/ofpstream.cc
kgbudge: Done.

src/c4/ofpstream.hh
kgbudge: Yes, because it allocates storage for its buffer and it does I/O, both of which are operations than can throw an exception.  I can add code to trap these, but I'd need to know how you would like these handled, if not as fatal exceptions.

src/c4/ofpstream.hh
kgbudge: Done.

kgbudge: Where should the DLL_PUBLIC_c4 go? Prefixing the class header?

src/c4/ofpstream.hh
kgbudge: Done.

src/c4/ofpstream.hh
kgbudge: Fixed by hand. M-q seems to reformat the comment to be longer than 80 in my current environment.

src/c4/ofpstream.hh
KineticTheory: @alexrlongne Do you have any thoughts on this?  Should we just allow the thrown exception to be fatal?

KineticTheory: Here's some additional information about why throwing dtors are bad.  http://bin-login.name/ftp/pub/docs/programming_languages/cpp/cffective_cpp/MEC/MI11_FR.HTM

kgbudge: Another possibility is to have the destructor simply discard any unflushed output. If this is done, the destructor does nothing but free its buffer, and freeing memory is not supposed to throw an exception.

KineticTheory: @kgbudge That sounds like a good plan.  Can you implement the nothrow dtor option that simply discards the unflushed output?

kgbudge: Done.

PR #429, opened 2018-07-02
Ensure that Processor_Group.t.hh is installed in the include/c4 directory.

kgbudge:
### Background

Almost all Capsaicin regression tests failed this weekend. This was due to a build failure which in turn was due to an #include "c4/Processor_Group.i.hh" directive which did not find the file.

It turns out this file was moved to c4/Processor_Group.t.hh. In addition, the c4/CMakeLists.txt file does not install .t.hh files in the include/c4 directory, so merely changing the Capsaicin code to reference c4/Processor_Group.t.hh did not resolve the problem.

Capsaicin needs access to the definitions of template members of rtt_c4::Processor_Group so that it can instantiate them for Teuchos::ArrayView<double>, the preferred vector container in the Capsaicin library.

### Description of changes

c4/CMakeLists.txt was modified to install .t.hh files from c4 into include/c4.

### Status

* Reference:
* Checks
  * [ ] Travis CI checks pass

### DIFF
--- a/src/c4/CMakeLists.txt
+++ b/src/c4/CMakeLists.txt
@@ -60,7 +60,8 @@ endif()
 # Installation instructions
 # ---------------------------------------------------------------------------- #
 install( TARGETS Lib_c4 EXPORT draco-targets DESTINATION ${DBSCFGDIR}lib )
-install( FILES ${headers} DESTINATION ${DBSCFGDIR}include/c4 )
+install( FILES ${headers} ${template_implementations} DESTINATION
+  ${DBSCFGDIR}include/c4 )
 
 # ---------------------------------------------------------------------------- #
 # Unit tests


KineticTheory:
@kgbudge @jhchang-lanl I'm going to push back on this a bit.  I think a better solution will be to:
 
(1) remove `c4/Processor_Group_pt.cc`,
(2) re-rename `c4/Processor_Group.t.hh` to `c4/Processor_Group.i.hh`, and
(3) modify `c4/Processor_Group.hh` so that it has an include directive at the end of the file for `c4/Processor_Group.i.hh` (use `C4_MPI.hh` as a guide).

As explained in the [Draco Coding Standards](https://rtt.lanl.gov/redmine/projects/draco/wiki/Coding_Standards), `.t.hh` files contain definitions for template functions that will be explicitly instantiated (and thus fully defined in the package library and preventing us from ever installing a `.t.hh` file).  In contrast, `.i.hh` files should be used for definitions of template functions that are instantiated implicitly.  Because Draco cannot (currently) provide an explicit instantiation for `void Processor_Group::sum(Teuchos::ArrayView &)`, it will be better to define this function implicitly.  Obviously, Capsaicin can provide an explicit instantiation at any time to decrease compile times or to restrict containers that `sum` and `assemble_vector` are allowed to operate on.

I've created #430 as a proposed alternate solution to this pull request.

PR #477, opened 2018-09-18
Restore the prune function to Ordinate_Space.

kgbudge:
The prune function of Ordinate_Space always returns true, and on this basis it was removed from the class. However, this is a virtual function that is nontrivially overridden by Galerkin_Ordinate_Space, and its removal changes the solution slightly in one of the Capsaicin benchmarks.

The difference is quite small, but until I can better recall why this was done in the first place, I suggest we restore the Ordinate_Space::prune function and restore the if-test using prune.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass

### DIFF
--- a/src/quadrature/Ordinate_Space.cc
+++ b/src/quadrature/Ordinate_Space.cc
@@ -322,9 +322,7 @@ void Ordinate_Space::compute_moments_(Quadrature_Class const quadrature_class,
     number_of_moments_ = 0;
     for (unsigned n = 0; n < moments_.size(); ++n) {
       int const l = moments_[n].L();
-      // member function prune() always returns true?
-      // if (l <= Lmax || !prune()) {
-      if (l <= Lmax) {
+      if (l <= Lmax || !prune()) {
         if (l > Lmax) {
           Lmax = l;
           moments_per_order_.resize(Lmax + 1, 0U);
--- a/src/quadrature/Ordinate_Space.cc
+++ b/src/quadrature/Ordinate_Space.cc
@@ -322,7 +322,7 @@ void Ordinate_Space::compute_moments_(Quadrature_Class const quadrature_class,
     number_of_moments_ = 0;
     for (unsigned n = 0; n < moments_.size(); ++n) {
       int const l = moments_[n].L();
-      if (l <= Lmax || !prune()) {
+      if (!prune() || l <= Lmax) {
         if (l > Lmax) {
           Lmax = l;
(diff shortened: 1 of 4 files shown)


KineticTheory:
@kgbudge There were testing issues on ccscs2 (not your fault).  I'll restart the tests.

src/quadrature/Ordinate_Space.cc
```diff
@@ -322,9 +322,7 @@ void Ordinate_Space::compute_moments_(Quadrature_Class const quadrature_class,
     number_of_moments_ = 0;
     for (unsigned n = 0; n < moments_.size(); ++n) {
       int const l = moments_[n].L();
-      // member function prune() always returns true?
-      // if (l <= Lmax || !prune()) {
-      if (l <= Lmax) {
+      if (l <= Lmax || !prune()) {
         if (l > Lmax) {
           Lmax = l;
```
KineticTheory: Can you provide a test case that covers this branch for this block of code?

```
        if (l > Lmax) {
          Lmax = l;
          moments_per_order_.resize(Lmax + 1, 0U);
        }
```

This seems like a corner case and I don't understand what conditions would lead us here.

src/quadrature/Ordinate_Space.hh
```diff
@@ -215,13 +215,11 @@ public:
...
-  //   // By default, prune any moments beyond the user-specified expansion
-  //   // order. Such moments are included in Galerkin methods for purposes of
-  //   // computing the M and D matrices, but are then removed from the moment
-  //   // space unless the GQF interpolation model has been specified.
-  // }
+  virtual bool prune() const {
+    return true;
+    // By default, prune any moments beyond the user-specified expansion order.
+    // Overridden by Galerkin_Ordinate_Space::prune().
+  }
```
KineticTheory: There is no test for this bit of new code.  Can you generate something to cover this case?

Ref: https://codecov.io/gh/lanl/Draco/compare/a118ab1cbbd7a27757cd6bc99e04a81aa1d6abd7...7d9dc881f45c58711e06085b5975d8f8736d265c/diff

KineticTheory:
@kgbudge It looks like the added function `virtual bool prune() const` from `Ordinate_Space.hh` remains uncovered.  Can you make it pure virtual or add a test?

src/quadrature/Ordinate_Space.hh
KineticTheory: Fixed by 0361632

PR #485, opened 2018-09-26
Avoid compiling for AVX2 instructions on older hardware.

KineticTheory:
### Background

+ The compile option `/arch:AVX2` was always enabled for Visual Studio, but pre-Haswell hardware does not support AVX2.  While the compiler could still create code, it would fail to run properly. To avoid this, we add a platform check for AVX2 capabilities -- actually the check already existed in `config/query_fma.cc` but it needed a small change to narrow the check to AVX2 only.

### Description of changes

+ Modify `query_fma.cc` to take command line arguments `-f` and `-a`, which limit the check to fma-availability or avx2-availablity only. This code will return `1` if those features are found.
+ Update `platform_checks.cmake` to compile and run `query_fma.cc` and then set the build variables `HAVE_HARDWARE_AVX2` and `HAVE_HARDWARE_FMA`. These values are saved to `ds++/config.h` and are also available when the build system parses `windows-cl.cmake` so that appropriate code paths can be taken.
+ Also:
  + Fix a comment block in `drel.f90`.
  + Fix a typo in `appveyor.yml` that prevented output from failing tests from being printed.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,7 @@ build_script:
 # no MPI installed so force scalar builds.
   - cmake -G "Visual Studio 15 2017" -DDRACO_C4=SCALAR %APPVEYOR_BUILD_FOLDER%
   - cmake --build . --config Debug
-  - ctest -C Debug -j 2 --output-on-failures
+  - ctest -C Debug -j 2 --output-on-failure
 
 #build:
 #  verbosity: minimal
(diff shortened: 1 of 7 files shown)


KineticTheory:
Compile fails on Cray.  Looking into it.

alexrlongne:
Looks good, I'm removing the WIP tag and merging after talking with @KineticTheory .

PR #497, opened 2018-10-04
Add a file include capability to Token_Streams.

kgbudge:
This capability allows file-based token streams to recognize a #include "filename" construct in their input and use it to include the contents of the named file in their token stream, analogous with the #include directive of the C preprocessor.

This initial implementation searches only the working directory for files specified with a relative path.

### Background

Test drivers often parse input files using the Token_Stream family of classes. Collections of input files, e.g. for test suites, often have considerable duplication of large parts of their input specifications. We can save a lot of work, especially maintenance of test suites when input formats change, by allowing input files to include other input files in the way that C/C++ programs include header files.

### Purpose of Pull Request

This pull request installs a first implementation of an include capability in the file-based token streams, File_Token_Stream and Parallel_File_Token_Stream. No include capability is provided yet for String_Token_Stream or Console_Token_Stream. Also, relative file paths are searched only in the current working directory.  Notwithstanding these limitations, I expect this to be a widely appreciated new capability.

### Description of changes

Any file containing a #include "filename" will be tokenized precisely as if the contents of filename were inserted in place of #include "filename". This is entirely analogous to the #include construct of the C preprocessor, except that the ability to search directories other than the working directory for the included file has not been implemented. 

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass


kgbudge:
K.T., can you take a look at the build error? I can't understand it at all. Builds fine on our cluster machines.

KineticTheory:
@kgbudge I'm looking at it.

kgbudge:
Thx


Kent G. Budge

CCS-2, LANL

7-9138


"Even duct tape can't fix stupid, but it can muffle the sound."

________________________________

KineticTheory:
@kgbudge `setenv` is not part of the C or C++ language standards and won't work on native-mac, [native-windows](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=vs-2017) or newer operating systems (newer windows and apple systems that run on phones, etc.).  If you must have this feature we will need to create platform specific versions in `ds++/SystemCall.hh`.

kgbudge:
I give up. This capability, while it would be very nice, is not essential, and there appears to be no portable way to set it up so it can be tested with our current testing system. Pulling the merge.

KineticTheory:
@kgbudge I'll look at this.  The cmake doesn't like the test name.  Maybe I didn't understand the documentation correctly.

src/parser/test/CMakeLists.txt
```diff
@@ -69,6 +69,16 @@ add_parallel_tests(
    PE_LIST "1;2"
    DEPS    "Lib_parser;${MPI_CXX_LIBRARIES};${PAPI_LIBRARY}" )
 
+# To ensure that the file token stream tests of the #include capability know
+# where to look for the included file.
+set( tests_with_env_reqs
```
KineticTheory: This breaks when we build without MPI because the test names are different w/o MPI.  Two of the CI tests build w/o MPI. Try this instead:

```
 if( "${DRACO_C4}" STREQUAL "MPI" )
   set( tests_with_env_reqs
     parser_tstParallel_File_Token_Stream_1
     parser_tstParallel_File_Token_Stream_2
     parser_tstFile_Token_Stream_1 )
 else()
   set( tests_with_env_reqs
     parser_tstParallel_File_Token_Stream
     parser_tstFile_Token_Stream )
 endif()
 set_tests_properties( ${tests_with_env_reqs} PROPERTIES
   ENVIRONMENT
   DRACO_INCLUDE_PATH=${CMAKE_CURRENT_SOURCE_DIR})
```

Note that I also removed the double quotes around ${CMAKE_CURRENT_SOURCE_DIR} to prevent the quotes from appearing as part of the variable's content after `getenv` is called.

KineticTheory:
This looks good to me.  I'll need to check the CDash report before I merge it.

PR #513, opened 2018-10-25
Make minor modifications following further usage of Draco_Mesh.

RyanWollaeger:
### Background

* In `Draco_Mesh`, side-node connectivity is only accessible indirectly through the cell-side layout data type.
* It is advantageous in certain contexts to have side-node linkage data directly accessible in serial form.
  * For instance, when integrating over a boundary, a simpler loop over the side index can be performed, without accessing an unneeded cell index per side.

### Purpose of Pull Request

* Make serial form of side-node linkage a data member of `Draco_Mesh`. 
* Also, add a virtual the `read_mesh` function to `Draco_Mesh_Reader` to enforce its existence in derived mesh readers.
* [Concerns Redmine Issue #1182](https://rtt.lanl.gov/redmine/issues/1182)

### Description of changes

+ Add serial side-node linkage data as members to Draco_Mesh.
+ Virtualize read_mesh function in Draco_Mesh_Reader base class.
+ Update unit test to test accessors to side-node linkage data.

### Status

* Reference:
* Checks
  * [ ] Travis CI checks pass


src/mesh/Draco_Mesh.hh
```diff
@@ -134,6 +137,12 @@ public:
   const std::vector<unsigned> &get_cell_to_node_linkage() const {
     return m_cell_to_node_linkage;
   }
+  const std::vector<unsigned> &get_side_node_count() const {
```
KineticTheory: You are returning a reference to member data. If the class goes out of scope the reference will be invalidated.  Is this the design you want?  (It is fine -- as long as you understand the possible downside).

src/mesh/test/tstDraco_Mesh.cc
```diff
@@ -65,6 +65,12 @@ void cartesian_mesh_2d(rtt_c4::ParallelUnitTest &ut) {
   if (mesh->get_cell_to_node_linkage() != cell_to_node_linkage)
     ITFAILS;
 
+  // check that flat side type and side to node linkage is correct
```
KineticTheory: Optionally - consider using `FAIL_IF_NOT(mesh->get_side_node_count() == side_node_count);`

src/mesh/Draco_Mesh.hh
RyanWollaeger: Thank you for the great question - so far the data has only been accessed when the class is in scope.  When the mesh is in scope, it may be expedient to access by reference.  But so far, when these vectors are accessed they're usual manipulated eventually and copied to another data structure.

src/mesh/test/tstDraco_Mesh.cc
RyanWollaeger: Thank you for the recommendation - I will plan to change these to `FAIL_IF_NOT`.

PR #555, opened 2018-12-20
Modify/Upgrade how the FMA CPP macro works.

KineticTheory:
### Background

* Jayenne uses the FMA macro to select accurate vs fast fused-multiply-add operations.  This allows us to use a high precision software version when hardware FMA are not available.

### Purpose of Pull Request

* [Related to Redmine Issue #1363](https://rtt.lanl.gov/redmine/issues/1363)
* [Related to Redmine Issue #1364](https://rtt.lanl.gov/redmine/issues/1364)

### Description of changes

* Eliminate the `FP_ACCURATE_FMA` mode for machines w/o hardware FMA.  Replace this with the CPP macro `FMA_ACCURATE`.
* For machines with hardware FMA, use the software call if it is faster, otherwise send `a*b+c` to the compiler
* For machines w/o hardware FMA, only use the very slow software call when `FMA_ACCURATE` is requested.
* Provide a special developer toggle `FMA_FIND_DIFFS` that can be set to have every call to FMA compute values using the software call and by direct evaluation of `a*b+c`.  When this mode is active, a report will be printed if the two calculations have a relative difference greater than 1.0e-13.  This is very useful for finding code that is sensitive to round-off errors.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass

### DIFF
--- a/src/ds++/FMA.hh
+++ b/src/ds++/FMA.hh
@@ -145,7 +145,7 @@ inline double fma_with_diagnostics(double const a, double const b,
 #define FMA_ACCURATE(a, b, c) fma((a), (b), (c))
 #else
 #define FMA(a, b, c) ((a) * (b) + c)
-#define FMA_ACURATE(a, b, c) ((a) * (b) + c)
+#define FMA_ACCURATE(a, b, c) ((a) * (b) + c)
 #endif
 
 #else /* HAVE_HARDWARE_FMA is false */
--- a/src/ds++/FMA.hh
+++ b/src/ds++/FMA.hh
@@ -4,7 +4,7 @@
  * \author  Kelly Thompson
  * \date    Thursday, Feb 09, 2017, 11:22 am
  * \brief   Provide extra control for FMA operations.
- * \note    Copyright (C) 2017-2018 Los Alamos National Security, LLC.
+ * \note    Copyright (C) 2017-2019 Los Alamos National Security, LLC.
  *          All rights reserved.
  *
  * Intel Haswell and later (and also modern AMD cpus) have hardware FMA
@@ -19,7 +19,7 @@
  * https://software.intel.com/en-us/node/405250?language=es&wapkw=avx2+cpuid
  *
  * \note The Intel compiler doesn't seem to set \c FP_FAST_FMA. If you know that
- * your algorithm is sensitive to roundoff, you might to locally disable
+ * your algorithm is sensitive to roundoff, you might need to locally disable
  * '-fp-model fast' by inserting this bit of code at the start of each function
  * that is sensitive:
  *
@@ -137,8 +137,8 @@ inline double fma_with_diagnostics(double const a, double const b,
  * compiler, so Intel will always pick the 2nd option.
  *
  * The intent of the \c FMA_ACCURATE version is to pick the more accurate
- * multiply-plus-add operation. When hardware FMA is available, we just just the
- * normal 'fast' version.
+ * multiply-plus-add operation. When hardware FMA is available, we just choose
+ * the normal 'fast' version.
  */
 #ifdef FP_FAST_FMA
 #define FMA(a, b, c) fma((a), (b), (c))
--- a/src/ds++/FMA.hh
+++ b/src/ds++/FMA.hh
@@ -144,7 +144,7 @@ inline double fma_with_diagnostics(double const a, double const b,
 #define FMA(a, b, c) fma((a), (b), (c))
 #define FMA_ACCURATE(a, b, c) fma((a), (b), (c))
 #else
-#define FMA(a, b, c) fma((a), (b), (c))
+#define FMA(a, b, c) ((a) * (b) + c)
 #define FMA_ACURATE(a, b, c) ((a) * (b) + c)
 #endif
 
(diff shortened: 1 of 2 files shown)


src/ds++/FMA.hh
```diff
@@ -7,16 +7,34 @@
...
+ * However, \c fma can be slow (maybe 3x slower).
  *
  * To improve solution consistency across platforms, we can choose to use a
  * software-based infinite precision fma call.
  *
  * \note Intel's code for detecting FMA availability on hardware.
  * https://software.intel.com/en-us/node/405250?language=es&wapkw=avx2+cpuid
+ *
+ * \note The Intel compiler doesn't seem to set \c FP_FAST_FMA. If you know that
+ * your algorithm is sensitive to roundoff, you might to locally disable
```
clevelam: might to locally -> might need to locally

src/ds++/FMA.hh
```diff
@@ -26,41 +44,129 @@
...
+ * Working on LANL systems, it seems that \c FP_FAST_FMA is not set by the Intel
+ * compiler, so Intel will always pick the 2nd option.
+ *
+ * The intent of the \c FMA_ACCURATE version is to pick the more accurate
+ * multiply-plus-add operation. When hardware FMA is available, we just just the
+ * normal 'fast' version.
  */
 #ifdef FP_FAST_FMA
-#define FMA(a, b, c) fma(a, b, c)
+#define FMA(a, b, c) fma((a), (b), (c))
```
clevelam: It seems to me that this ifdef is switchted. Isn't ((a)*(b)+(c)) always the fastest?

clevelam: Also, why do we need FMA and FMA_ACCURATE, it looks like there definitions are always the same in these if blocks.

src/ds++/FMA.hh
```diff
@@ -26,41 +44,129 @@
...
+ * This will be the default for new machines. Both versions should be equally
+ * accurate, choose whatever is faster. \c FP_FAST_FMA is set by the compiler
  *
  * \ref http://en.cppreference.com/w/c/numeric/math/fma
+ *
+ * Working on LANL systems, it seems that \c FP_FAST_FMA is not set by the Intel
+ * compiler, so Intel will always pick the 2nd option.
+ *
+ * The intent of the \c FMA_ACCURATE version is to pick the more accurate
+ * multiply-plus-add operation. When hardware FMA is available, we just just the
```
clevelam: just just -> just

src/ds++/FMA.hh
```diff
@@ -26,41 +44,129 @@
...
+ * that will call an infinite precision fma function so that numerical
+ * reproducibility is enhanced This allows solutions generated on older machines
+ * to match solutions generated on newer machines.
+ *
+ * Operations that are known to be very sensitive to round-off error should use
+ * \c FMA_ACCURATE even though it is much slower (2-3x slower).
+ */
 #define FMA(a, b, c) ((a) * (b) + c)
-#endif
+#define FMA_ACCURATE(a, b, c) fma((a), (b), (c))
```
clevelam: I see, this is where the FMA_ACCURATE and FMA are different.

clevelam: Is this going to bite us if we fail to pickup the native FMA on the platform?

src/ds++/FMA.hh
KineticTheory: Thanks for catching this.

src/ds++/FMA.hh
KineticTheory: * If `FP_FAST_FMA` is set, then the call to `fma(a,b,c)` is guaranteed to be the fastest computation (yes - faster than `a*b+c` and yes - it is confusing).  You can read the section at http://en.cppreference.com/w/c/numeric/math/fma for more details.
* I want the CPP macro `FMA_ACCURATE` to always be defined.  This allows us to always choose the more _accurate_ version, even when it is slower.  This is required in a few places to allow our tests to pass.  For machines that have hardware FMA, `FMA_ACCURATE == FMA`, but this isn't true for machines w/o hardware FMA.

src/ds++/FMA.hh
KineticTheory: fixed.

src/ds++/FMA.hh
KineticTheory: * Right
* We will deal with that when we encounter it.  As always, the problem is reproducibility of solutions between platforms and compilers and compiler options.  We could always choose the 'fastest', but then our test setup is worthless.

KineticTheory: I'm sure we will be revisiting this whole issue later this year due to porting to GPU and ARM.

src/ds++/FMA.hh
```diff
@@ -26,41 +44,129 @@
...
+ * multiply-plus-add operation. When hardware FMA is available, we just just the
+ * normal 'fast' version.
  */
 #ifdef FP_FAST_FMA
-#define FMA(a, b, c) fma(a, b, c)
+#define FMA(a, b, c) fma((a), (b), (c))
+#define FMA_ACCURATE(a, b, c) fma((a), (b), (c))
 #else
 #define FMA(a, b, c) ((a) * (b) + c)
+#define FMA_ACURATE(a, b, c) ((a) * (b) + c)
```
keadyk: It seems "accurate" is misspelled here!

KineticTheory: Fixed.  Thanks for catching that one.

keadyk:
Got it! :)

PR #564, opened 2019-01-16
Experimental simplified class parser templates.

kgbudge:
### Background

The parser directory has a Class_Parse_Table template and associated supporting boilerplate for writing parsers to read a text file description of how to create a class object. This template is a bit clunky, with more in it than is needed (such as options for partial or case-insensitive keyword matching, which we have never used.) It also relies on static functions for parsing, with the need to manage an artificial global "this" pointer for the object in which parsed data is stored. This can be made reentrant but only with extra effort.

### Purpose of Pull Request

Add a new Class_Parser template that tries to bring everything into one place and streamline it.  The biggest change is to use pointers to member functions in the keyword tables as parse functions. This eliminates the artificial "this" pointer, simplifying code and making the parser automatically reentrant, but comes at the cost that it precludes diamond inheritance of class parsers. (Because a pointer to base class member function can be cast to pointer to derived class member function only if the base cass is nonvirtual.)

WIP: Offered for comment. Is this really a significant improvement?

### Status

* Reference:
* Checks
  * [ ] Travis CI checks pass

### DIFF
--- a/src/parser/Class_Parser.hh
+++ b/src/parser/Class_Parser.hh
@@ -1,0 +1,411 @@
+//----------------------------------*-C++-*-----------------------------------//
+/*!
+ * \file   parser/Class_Parser.hh
+ * \brief  Definition of Class_Parser_Keyword and Class_Parser.
+ * \note   Copyright (C) 2016-2019 TRIAD, LLC. All rights reserved. */
+//----------------------------------------------------------------------------//
+
+#ifndef rtt_Class_Parser_HH
+#define rtt_Class_Parser_HH
+
+#include "Token_Stream.hh"
+#include <cstring> // strcmp
+#include <memory>
+#include <vector>
+
+namespace rtt_parser {
+
+//----------------------------------------------------------------------------//
+/*! Template for parser table classes
+ *
+ * A Class_Parser is conceptually a table of keywords and associated parsing
+ * functions. When invoked (via a call to the Class_Parser::parse() member
+ * function), it pulls tokens from a Token_Stream, and attempts to match the
+ * tokens against a table of keywords.  Each keyword has a parsing function and
+ * integer flag associate with it, and when a keyword matches an input token,
+ * the parsing function is called with the original Token_Stream and the integer
+ * flag as arguments. The parsing functions can pull additional tokens from the
+ * Token_Stream, such as parameter values, prior to returning control to the
+ * Class_Parser. The machinery to do all this is normally provided by deriving
+ * the Class_Parser from Class_Parser_Base.
+ *
+ * Each parse function is a member function of the Class_Parser, and the
+ * values it parses are normally stored in the Class_Parser for later use
+ * to construct an object of the class for which the Class_Parser is
+ * specialized. A pointer to this object is returned by parse().
+ *
+ * When the Class_Parse_Table_Base fails to match an input keyword to its
+ * keyword table, or when the input token is not a keyword, END, EXIT,
+ * or ERROR, the parse table reports an error, then attempts recovery
+ * by reading additional tokens until it encounters either a keyword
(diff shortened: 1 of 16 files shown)


src/parser/Class_Parser.hh
```diff
@@ -0,0 +1,411 @@
...
+ *
+ * In addition to these mandatory members, any nontrivial Class_Parser
+ * will also define data members to hold the parsed specifications, and parse
+ * functions of the form
+ *
+ * void parse_parameter(Token_Stream &, int);
+ *
+ * A nontrivial Class_Parser should define a member
+ *
+ * static Class_Parser_Keyword<Class> const raw_table[];
```
KineticTheory: You will need to escape the angle brackets in the documentation block.  Doxygen thinks you are inserting html commands:

```
Class_Parser_Keyword \<Class\>
```

Please fix here and below.

src/parser/Class_Parser.ii
```diff
@@ -0,0 +1,658 @@
+//----------------------------------*-C++-*-----------------------------------//
+/*!
+ * \file   Class_Parser.ii
```
KineticTheory: This is a non standard file extension (at least for Draco).  Why are you introducing it?  How is it different from existing file extensions?

src/parser/Class_Parser.ii
```diff
@@ -0,0 +1,658 @@
...
+
+    sort_table_();
+  }
+
+  Ensure(check_class_invariants());
+}
+
+#if 0
+//----------------------------------------------------------------------------//
+/*!
```
KineticTheory: Can you just remove this 'dead' code block?

src/parser/Class_Parser.hh
kgbudge: I'll try wrapping this section in \code and see if it makes it better.

src/parser/Class_Parser.ii
kgbudge: Are you no longer using .ii for implementation details of templates? Should I move the implementation into the .hh file?

KineticTheory: We use `.i.hh` for implementation details (to separate implementation from the design captured in the .hh` file.  Include the `.i.hh` form the bottom of the `.hh` file).  Use `.t.hh` for explicit template instantiation.  Ref: https://rtt.lanl.gov/redmine/projects/draco/wiki/Coding_Standards

kgbudge: Ah. Will do.

src/parser/Class_Parser.hh
```diff
@@ -0,0 +1,417 @@
...
+ * will inspect this table for errors and make a sorted copy for fast token
+ * matching.
+ *
+ * The specialization will thus look something like
+ *
+ * Class__parser.hh:
+ *
+ * \code
+ * template<>
+ * class Class_Parser<Class> : public Class_Parser_Base<Class>
```
KineticTheory: code formatting check is failing.  See https://github.com/lanl/Draco/wiki/Check-Fix-Code-Style-Issues

src/parser/Class_Parser.hh
```diff
@@ -0,0 +1,417 @@
...
+
+#include "Token_Stream.hh"
+#include <cstring> // strcmp
+#include <memory>
+#include <vector>
+
+namespace rtt_parser {
+
+//----------------------------------------------------------------------------//
+/*! Template for parser table classes
```
KineticTheory: Autodoc tests are failing:

```
Gener/home/travis/Draco/src/parser/Class_Parser.hh:325: error: The following parameters of rtt_parser::Class_Parser_Base::Class_Parser_Base(Child &child, Class_Parser_Keyword< Class > const *raw_table, unsigned const count) are not documented:
  parameter 'child' (warning treated as error, aborting now)
```

PR #571, opened 2019-02-12
Re-enable compilation with CUDA.

KineticTheory:
### Background

+ CUDA and CMake have changed a lot since the original implementation of the code in the `device` package. This change-set is still a bit clunky, but I wanted to get the essentials updated in our build system so other developers can start looking at CUDA.
  - CUDA is now a first class language in CMake.  We will treat it like Fortran (an optional compiler).
+ Testing of these changes was done on Darwin's `volta-x86` node with `gcc/7.3` and `cuda/9.2`  The bashrc file for Darwin was updated to try to do the right thing for the `volta-x86` nodes.
+ Note: Nvidia is now discouraging the use of the `driver-API`. Much of the original implementation used this API.  This code is left our repository for reference, but isn't actually compiled at this time.

### Purpose of Pull Request

* [Related to Redmine Issue #1168](https://rtt.lanl.gov/redmine/issues/1168)

### Description of changes

+ Register CUDA as an optional compiler language that will be enabled in the Draco build system if `nvcc` is found in the path.  The build system will use the boolean variable `WITH_CUDA` to control when cuda code is considered for compilation.  Ref: top-level `CMakeLists.txt` and `config/vendor_libraries.cmake`.
+ Comment out (but did not yet remove) a lot of the old CUDA build system code.
+ Deprecate the `USE_CUDA` variable.  Use `WITH_CUDA` instead, as this is more standard.
+ The `GPU_Device` code was updated to replace the use of deprecated CUDA commands.  In truth, this class needs to be updated to use the C runtime API for cuda instead of the device-API before the class is useful again.
+ More tests need to be added as examples.  I think we can take some of these from the acceleware training or cmake+cuda tutorials.
+ Environment changes (`.bashrc_darwin`)
  - When on Darwin's `volta-x86` node automatically load the `cuda/9.2` module.
  - When any Darwin x86 node, add `clang-format` (v.6) to the developer's PATH.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass


KineticTheory:
Wow - the CI testing is an abysmal failure.  I'll look into it. It might be related to changes to Travis and not directly related to this PR.

src/device/test/GPU_Device.cmake
```diff
@@ -1,124 +1,83 @@
 #-----------------------------*-cmake-*----------------------------------------#
```
RyanWollaeger: Wow - this is an impressive clean-up!

RyanWollaeger:
It looks like the CDash builds are coming in clean.

src/device/GPU_Device.hh
```diff
@@ -41,7 +42,7 @@ namespace rtt_device {
  *         << "\n" << endl;
  *     for( size_t device=0; device<numDev; ++device )
  *         gpu.printDeviceSummary( device, out );
- *   
+ *
  *   // Print the message to stdout
  *   cout << out.str();
  * \endcode
```
RyanWollaeger: It appears there is an autodoc error for this file: `Preprocessing /home/home/travis/Draco/src/device/GPU_Device.hh:52: error: included file device/test/gpu_hello_rt_api.cc is not found. Check your EXAMPLE_PATH (warning treated as error, aborting now)`.  Should this example file be suffixed with `.cu`?

src/device/GPU_Device.cc
```diff
@@ -57,10 +60,75 @@ GPU_Device::GPU_Device(void) // int /*argc*/, char */*argv*/[] )
...
-    checkForCudaError(err);
-    deviceProperties.push_back(properties);
+    {
+      int tmp1(0), tmp2(0), tmp3(0);
+      err = cuDeviceGetAttribute(
+          &tmp1, CU_DEVICE_ATTRIBUTE_MAX_THREADS_PER_BLOCK, cuDevice);
+      checkForCudaError(err);
+      m_maxthreadsperblock.push_back(tmp1);
+
+      err = cuDeviceGetAttribute(&tmp1, CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_X,
```
RyanWollaeger: Very nice - I don't think we covered these attribute accessors in the training.

src/device/GPU_Device.hh
KineticTheory: I fixed this.  I also added a real call to a cuda kernel in the example/test.

KineticTheory:
@RyanWollaeger Finally a clean CI.

src/device/GPU_Device.hh
RyanWollaeger: The test looks great!

RyanWollaeger:
@KineticTheory Is this still WIP?

KineticTheory:
I just removed the `WIP:`

PR #574, opened 2019-02-18
Move LaTeX related files to the dracodoc repository

KineticTheory:
### Background

+ Better to collocate these files in the same repository as the actual documentation.


KineticTheory:
@jhchang-lanl I ripped out all of the LaTeX files from the main draco repository and moved these files to dracodoc and copied them into imcdoc.  Does your project need any of these files?  If not, I will go ahead and merge this PR.

PR #592, opened 2019-04-03
Feature add atomics

timmah:
### Background

* Add atomic support for floating point types. This will be obsoleted when Draco adopts C++20.

### Purpose of Pull Request

* add support for atomic fetch-and-add for floating point types
* it works, can't say it's great performance

### Description of changes

* Add ds++/atomics.hh (must...not...call it house_atomics.hh)
* Add multithreaded unit test in ds++/test/tstatomics.cc that attempts to demonstrate correct use of atomics
* Requires C++11 <atomic> support

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass


src/ds++/DracoStrings.hh
```diff
@@ -115,6 +115,9 @@ template <typename T> auto parse_number_impl(std::string const &str) -> T;
 // specializations for these types are defined in DracoStrings.cc
 template <> auto parse_number_impl<int32_t>(std::string const &str) -> int32_t;
 template <> auto parse_number_impl<int64_t>(std::string const &str) -> int64_t;
+template <> auto parse_number_impl<long>(std::string const &str) -> long;
```
timmah: Somehow I got into trouble not having these declarations. Trying to remember where they were needed; if I can't, will delete them.

src/ds++/atomics.hh
```diff
@@ -0,0 +1,71 @@
...
+    to_store = arg - expected;
+  }
+  return to_store;
+}
+
+} // namespace rtt_dsxx
+
+#endif
+
+// End of file
```
KineticTheory: Please use the standard footer block
```
#endif // __dsxx_Atomics_hh__

// end of ds++/atomics.hh
```

src/ds++/test/tstatomics.cc
```diff
@@ -0,0 +1,165 @@
+//----------------------------------*-C++-*----------------------------------//
+/*!
+ * \file   ds++/test/tstatomics.cc
+ * \author Tim Kelley
+ * \date   Thursday, Sept. 6, 2018, 10:51 am
+ * \note   Copyright (C) 2018 Los Alamos National Security, LLC.
```
KineticTheory: fix copyright line

src/ds++/test/tstatomics.cc
```diff
@@ -0,0 +1,165 @@
...
+    u.passes(msg);
+  }
+  return;
+}
+
+//----------------------------------------------------------------------------//
+int main(int argc, char *argv[]) {
+  rtt_dsxx::ScalarUnitTest ut(argc, argv, rtt_dsxx::release);
+  try {
+    run_a_test(ut, test_fetch_add_atomic, "fetch_add ok.");
```
KineticTheory: Nice test!

src/ds++/atomics.hh
```diff
@@ -0,0 +1,71 @@
...
+namespace rtt_dsxx {
+
+/**\brief atomically add arg to a.
+ * \tparam FpT: a floating point type (integer types in std lib since C++11).
+ * \param a: an atomic of type FpT that will be updated.
+ * \param arg: quantity to add to a
+ * \param m_o: std memory order, default is memory_order_relaxed
+ * \return value of a after update
+ * \remark: By default, uses memory_order_relaxed, meaning (I think) that other
+ * atomic operations on 'a' can be moved before or after this one.
```
alexrlongne: Yeah, that's my reading too. I think this is what we want for our default behavior. I like that it's a default argument. If for some reason we want to get into ordering for reproducibility we could make this some kind of configurable thing like `FMA` (I sincerely hope we don't have to do that).

alexrlongne:
@KineticTheory I agree with your suggestions, otherwise this looks good to me! I think it's great that the standard has this level of control for atomics. I don't quite understand why they don't provide these  kinds of basic operations on atomic types.

timmah:
I want to add a test of fetch_sub, maybe hold off before merging.

attom:
This looks reasonable to me. I assume there's no problem using compare_exchange_weak instead of strong?

PR #613, opened 2019-05-01
Fix issue 1268 by documenting and testing report_leaks.

kgbudge:
Take another stab at a portable test of rtt_memory::out_of_memory_handler. Also, bring this function into conformance with the standard, by having it set the memory handler to null and return rather than directly throw a bad_alloc.

### Background

Issue 1268 is that rtt_memory::report_leaks is neither documented nor tested. Coverage analysis also shows that rtt_memory::out_of_memory_handler is not tested. The line to install this handler is commented out of the unit test program.

The current implementation of rtt_memory::out_of_memory_handler does not conform to the standard, since it throws a bad_alloc directly. The standard calls for this function to set the out of memory handler to nullptr if it cannot resolve the allocation failure, and then the new allocator knows to throw bad_alloc itself. 

### Purpose of Pull Request

* [Fixes Redmine Issue #1268](https://rtt.lanl.gov/redmine/issues/1268)

### Description of changes

* Add test of report_leaks to tstmemory.cc
* Rewrite out_of_memory_handler to conform to the standard.
* Reactivate the test of out_of_memory_handler so it is covered.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass

### DIFF
--- a/src/memory/test/tstmemory.cc
+++ b/src/memory/test/tstmemory.cc
@@ -20,6 +20,8 @@ using namespace rtt_memory;
 // TESTS
 //---------------------------------------------------------------------------//
 
+size_t get_really_big_size_t();
+
 void tst_memory(rtt_dsxx::UnitTest &ut) {
   if (total_allocation() == 0) {
     PASSMSG("correct initial total allocation");
@@ -59,11 +61,10 @@ void tst_memory(rtt_dsxx::UnitTest &ut) {
   } else {
     FAILMSG("NOT correct largest allocation");
   }
-
-  report_leaks(cerr);
 #else
   PASSMSG("memory diagnostics not checked for this build");
 #endif
+  report_leaks(cerr);
 
   delete[] array;
   delete[] array2;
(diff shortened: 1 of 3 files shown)


src/memory/test/tstmemory.cc
```diff
@@ -165,18 +162,22 @@ void tst_bad_alloc(rtt_dsxx::UnitTest &ut) {
...
 
   return;
 }
 
+//----------------------------------------------------------------------------------------//
+// Some compilers are clever enough to figure out that if you pass
+// std::numeric_limits<size_t>::max() to the new operator, you will always
+// blow away member, and so they will refuse to compile the code. We have
+// to use a bit of indirection to get such compilers to swallow the huge
+// allocation meant to deliberately blow away memory.
```
KineticTheory: Evidently, even this trick isn't enough.  

```
src/memory/test/tstmemory.cc:148:55: warning: argument 1 value ‘18446744073709551615’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
```

Can you try adding the following CPP macro to disable GCC's `-Walloc-size_larger-than` check?

```
#include <vector>

#if defined(__GNUC__) && !defined(__clang__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Walloc-size_larger-than"
#endif

<test code goes here>

#if defined(__GNUC__) && !defined(__clang__)
#pragma GCC diagnostic pop
#endif
```

KineticTheory:
@kgbudge BTW, the compile warnings about `tstSuperludist.cc` are a known issue that I will take care of in a separate PR.  You can ignore [such warnings.](https://rtt.lanl.gov/cdash/viewBuildError.php?type=1&buildid=25644)

src/memory/test/tstmemory.cc
kgbudge: Done.

KineticTheory:
@kgbudge Hmm. My recommended fix didn't solve the problem.  After a bit of research, it appears that you have stumbled upon [a known bug in gcc-8.1.0](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87544).

Can you rip out the `#pragma` stuff I just had you add and then try [Gael Guennebaud's recommended work around.](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87544#c2)  Something like this:

```c++
size_t get_really_big_size_t() {
  return numeric_limits<std::ptrdiff_t>::max() / sizeof(size_t);
}
```

PR #615, opened 2019-05-03
Fix build warnings related to superlu_ddefs.h.

KineticTheory:
### Background

+ Recently, gcc-8.1.0 regressions have started to generate the following compile warning:

```
superlu_dist_config.h:2:6: warning: "XSDK_INDEX_SIZE" is not defined, evaluates to 0 [-Wundef]
```

+ Since the issue isn't related to code that we own, I have created a
  suppression as described below.

### Purpose of Pull Request

* Fixes #614

### Description of changes

+ Wrap the inclusion of this header in a new file `superlu-dist-wrapper.h` and update the list of include directories to allow the build system to find the new file.
+ Tag `superlu_ddefs.h` with `pragma GCC system_header` to prevent GCC from creating warnings about code found in the TPL header.

### Status

* Reference:
  * [Pre-Merge Code Review](https://github.com/lanl/Draco/wiki/Style-Guide)
  * [CDash Status](https://rtt.lanl.gov/cdash/index.php?project=Draco&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=-pr)
* Checks
  * [x] Travis CI checks pass
  * [x] Code coverage does not decrease
  * [x] [Valgrind test passes](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [x] [Toss3 checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [x] [Trinitite checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] Code reviewed/approved, sufficient DbC checks, testing, documentation

PR #617, opened 2019-05-06
Cleanup warnings/errors reported on CDash.

KineticTheory:
### Background:

+ `VendorsChecks_tstSuperludist_4` has been failing for Win32 for a while.  There is a problem with the the Win32 build of superlu-dist.  I don't have time to fix this right now and this test isn't particularly important for Draco or Jayenne.
+ A new 'unused variable' warning showed up late last week (`tstmemory`).  This test also fails when `fpe_trap` is enabled.

### Changes:

+ Disable tstsuperludist for WIN32 platforms.
+ Comment out unused variable.
+ Disable `memory_tstmemory` when `DRACO_DIAGNOSTICS=7` (fpe_trap enabled).  This removes it from the _fulldiagnostics_ regression suite.
+ Also
  - For Cray systems, force `KMP_AFFINITY=disabled` via the developer bashrc files to avoid a thread affinity issue identified by Tim.
  - Add sanity checks to `push_repository_xf.sh` after Jae reported issues when trying to run this script.

### DIFF
--- a/environment/bashrc/.bashrc_cray
+++ b/environment/bashrc/.bashrc_cray
@@ -143,6 +143,9 @@ function dracoenv ()
 
 function rmdracoenv ()
 {
+  # this setting recommended by Sam Guiterrez to allow correct core binding for
+  # C++ threads (maybe OpeNMP threads also) on KNLs.
+  export KMP_AFFINITY=disabled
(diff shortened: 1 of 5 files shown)


KineticTheory:
@attom All test pass.  There are 3 build warnings -- but these are a known issue and will be fixed by #615.  This is ready for final review and merge.

attom:
LGTM

PR #619, opened 2019-05-13
nGray_Analytic_MultigroupOpacity to Compound_Analytic_MultigroupOpacity

kgbudge:
Likewise rename nGray_Analytic_Odfmg_Opacity to Compound_Analytic_Odfmg_Opacity.

Change Compound_Analytic_MultigroupOpacity to use group structure when calculating opacities, to make it consistent with Compound_Analytic_Odfmg_Opacity previously used in Capsaicin.

### Background

The Capsaicin project is, so far as we know, the only client for the OdfmgOpacity class hierarchy. This was written to support experiments on using opacity distribution functions for thermal transport calculations. The results were not encouraging and no further work has been done in this direction in years, nor do we foresee any further work in the future. Capsaicin has therefore abandoned support of opacity distribution functions.

This removes significant code bloat from Capsaicin, but falling back to MultigroupOpacity from Odfmg_Opacity revealed that the nGray_Analytic_MultigroupOpacity class does not produce the same opacities as nGray_Analytic_Odfmg_Opacity does for a single distribution band. Further investigation showed thatn nGray_Analytic_MultigroupOpacity ignores frequency dependence in its opacity calculation. 

It was not clear at first whether this was deliberate. Digging into the code convinced me that I had not intended nGray_Analytic_MultigroupOpacity to have the same opacity in all groups in all cases. The name seems confusing enough that Compound_Analytic_... seemed more appropriate to capture the concept, which is of an opacity built on a set of individual group models.

### Purpose of Pull Request

Give nGray ... classes a more appropriate name, Compound...
Make Compound_Analytic_MultigroupOpacity match Compound_Analytic_OdfmgOpacity for a single band per group.

### Description of changes

Renamed classes.
Modified opacity calculation loop to use the frequency dependent interface of the Analytic_Opacity_Model class.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass

### DIFF
--- a/src/cdi_analytic/Compound_Analytic_MultigroupOpacity.cc
+++ b/src/cdi_analytic/Compound_Analytic_MultigroupOpacity.cc
@@ -1,14 +1,12 @@
 //----------------------------------*-C++-*----------------------------------//
 /*!
- * \file   cdi_analytic/nGray_Analytic_MultigroupOpacity.cc
- * \author Thomas M. Evans
- * \date   Tue Nov 13 11:19:59 2001
- * \brief  nGray_Analytic_MultigroupOpacity class member definitions.
+ * \file   cdi_analytic/Compound_Analytic_MultigroupOpacity.cc
+ * \brief  Compound_Analytic_MultigroupOpacity class member definitions.
  * \note   Copyright (C) 2016-2019 Triad National Security, LLC.
  *         All rights reserved. */
 //---------------------------------------------------------------------------//
 
-#include "nGray_Analytic_MultigroupOpacity.hh"
+#include "Compound_Analytic_MultigroupOpacity.hh"
 #include "ds++/Packing_Utils.hh"
 #include "ds++/dbc.hh"
 
@@ -27,7 +25,7 @@ namespace rtt_cdi_analytic {
  * rtt_cdi::Reaction argument.
  *
  * The group structure (in keV) must be provided by the groups argument.  The
- * number of Analytic_Opacity_Model objects given in the models argument must 
+ * number of Analytic_Opacity_Model objects given in the models argument must
  * be equal to the number of groups.
  *
  * \param groups vector containing the group boundaries in keV from lowest to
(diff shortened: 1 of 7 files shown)


attom:
Everything looks good except for the naming, which I don't quite understand. You're renaming "nGray" because the opacities are not `G` copies of a gray opacity, right? But what makes them "Compound"? They're just a MG or MG+MB opacity, where there's a single formula used to figure out the opacity in all groups?

keadyk:
Sorry @KineticTheory, this totally slipped my mind! Reviewing now :)

kgbudge:
> Everything looks good except for the naming, which I don't quite understand. You're renaming "nGray" because the opacities are not `G` copies of a gray opacity, right? But what makes them "Compound"? They're just a MG or MG+MB opacity, where there's a single formula used to figure out the opacity in all groups?

Each group is linked to its own Analytic_Opacity_Model (though often it's the same model for all groups, via shared_ptr) and so the Compound_Analytic_MultigroupOpacity is "compounded" of Analytic_Opacity_Models.

src/cdi_analytic/Compound_Analytic_MultigroupOpacity.cc
```diff
@@ -120,28 +124,32 @@ nGray_Analytic_MultigroupOpacity::nGray_Analytic_MultigroupOpacity(
...
+  auto nu = getGroupBoundaries();
+  Check(nu.size() == group_models.size() + 1);
+
   for (size_t i = 0; i < opacities.size(); ++i) {
     Check(group_models[i]);
 
     // assign the opacity based on the group model
-    opacities[i] = group_models[i]->calculate_opacity(temperature, density);
+    opacities[i] = group_models[i]->calculate_opacity(temperature, density,
+                                                      nu[i], nu[i + 1]);
```
keadyk: Just to be sure I understand -- this update will cause a change in behavior if the group_model is Polynomial_Analytic_Opacity_Model, correct?

kgbudge: Yes. This is the change that brings the behavior into consistency with the ODF version.

keadyk: Excellent, thanks!

attom:
That makes sense. Looks good to me, then.

PR #630, opened 2019-06-06
Deprecate Insist.

kgbudge:
Add AlwaysRequire, AlwaysCheck, AlwaysEnsure that resemble Insist but are DBC concepts.

### Background

Insist was originally conceived as a quick and dirty way to report user error. Thus, it is always active. However, this is a poor way to diagnose user _input_ error, and it tends to confound user input error with contract violations (which are code errors, not user input errors.)

At the same time, I felt a need for equivalents of Require, Check, Assert that would always be on. These will be used for inexpensive checks in non-time-critical parts of the code, so that they can be left on in production code. If a production calculation fails, the user can then directly report the assertion message and location. These versions also always include a stack trace in the message. This should make diagnosis of user errors a bit faster, and I think there are psychological advantages for users seeing a tangible error report they can pass to us rather than an out of the blue "segmentation violation" or something equally obscure. The always-on versions are prefixed by Always, e.g. AlwaysRequire.

### Purpose of Pull Request

* Deprecate Insist
* Add AlwaysRequire, AlwaysCheck, and AlwaysEnsure macros.

### Description of changes

* Mark Insist as deprecated in its Doxygen documentation.

* Add AlwaysRequire, AlwaysCheck, and AlwaysEnsure macros that are used in the same contexts as Require, Check, and Ensure but are always on and always throw an assertion with a stack trace.

* Add a print_stack argument to assertion constructor and associated functions and modify the old Require, Check, Ensure macros to pass false. The new macros pass true. This preserves the existing behavior of the old macros under different builds while enabling the desired behavior of the new macros.

* Add test code to exercise the new macros.

### Status

* Reference:
* Checks
  * [ ] Travis CI checks pass


KineticTheory:
@kgbudge @alexrlongne @jhchang-lanl Since this is a fairly invasive change, I would like to postpone merging it until we finish the releases planned for early next week.  Comments?

kgbudge:
There is no rush at all on this.

KineticTheory:
@jhchang-lanl @alexrlongne We need a plan before this is merged since it will break both of our codes temporarily. Can you help prepare your codes for this change?

alexrlongne:
@kgbudge @KineticTheory I have a branch that makes the appropriate changes. I can push it whenever you're ready.

kgbudge:
Whatever became of this?

KineticTheory:
Outstanding issues before #630 can be merged:

* [ ] Fixes to this PR to allow the Travis checks to pass.
* [ ] Completed review from @jhchang-lanl (or his representative)
* [ ] Completed review from @alexrlongne (I think he has prepared our code for this PR.)

kgbudge:
Meh, there doesn't seem to be much interest in pushing this forward. Closing.

PR #646, opened 2019-07-03
Performance tweaks for Planck and Rosseland integration

alexrlongne:
### Background

* I've been looking into the cycle initialization and finalization times in Jayenne and Draco's various Planck and Rosseland integration functions show up as being fairly expensive in the profiler. I've tried to reduce the number of times they're called in Jayenne but I also noticed a few possible improvements to the integration routines themselves. The two main improvements are relaxing some input checking and making a comparison value `constexpr`. Currently, `sqrt(std::numeric_limits<double>::max()` is calculated every time a Planck integration routine is called. This value should be a `constexpr` but due to `sqrt` not being marked as `constexpr` the function is evaluated every time. I introduced a very simple version of `sqrt` for these kinds of cases into `ds++`.

* This PR also removes the unnecessary copying in some C4 calls.`MPI_IN_PLACE` takes care of this--I think there used to be a bug in Open_MPI related to this keyword but it's been fixed since around version 1.10.

### Purpose of Pull Request

* [Fixes Redmine Issue #1565](https://rtt.lanl.gov/redmine/issues/1565)

### Description of changes
* Introduce a Constexpr_Functions file to help define constant values
* Use MPI_IN_PLACE whenever possible in c4
* Allow for more stack depth in constexpr functions calls in gcc
### Status

* Reference:
* Checks
  * [x] Travis CI checks pass
  * [x] Test under Cray and XL compilers before merging.

### DIFF
--- a/src/cdi/CDI.cc
+++ b/src/cdi/CDI.cc
@@ -260,18 +260,16 @@ void CDI::integrate_Rosseland_Planckian_Spectrum(const size_t groupIndex,
 void CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
                                        double const T,
                                        std::vector<double> &planck) {
-  Require(T >= 0.0);
+  Require(T >= std::numeric_limits<double>::min() &&
+          T < std::numeric_limits<double>::max());
   Require(bounds.size() > 0);
 
   size_t const groups(bounds.size() - 1);
   planck.resize(groups, 0.0);
 
-  // nu/T must be < numeric_limits<double>::max().  So, if T < nu*min(), then
-  // return early with planck == 0.0. This avoids a possible divide by zero but
-  // it doesn't catch the case where bounds[0] == 0.0 and T is a denormal so
-  // make it the more restrictive T <= numeric_limits<double>::min()
-  if (T <= std::numeric_limits<double>::min())
-    return;
+  // nu/T must be < numeric_limits<double>::max(), throw assertion in DBC mode
+  // if this occurs
+  Check(T > bounds.back() * std::numeric_limits<double>::min());
 
   // Initialize the loop:
   double scaled_frequency = bounds[0] / T;
@@ -307,19 +305,17 @@ void CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
 std::vector<double>
 CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
                                   double const T) {
-  Require(T >= 0.0);
+
+  Require(T >= std::numeric_limits<double>::min() &&
+          T < std::numeric_limits<double>::max());
   Require(bounds.size() > 0);
 
   size_t const groups(bounds.size() - 1);
   std::vector<double> planck(groups, 0.0);
 
-  // nu/T must be < numeric_limits<double>::max().  So, if T < nu*min(), then
-  // return early with planck == 0.0. This avoids a possible divide by zero but
-  // it doesn't catch the case where bounds[0] == 0.0 and T is a denormal so
-  // make it the more restrictive T <= numeric_limits<double>::min()
-
-  if (T <= std::numeric_limits<double>::min())
-    return planck;
+  // nu/T must be < numeric_limits<double>::max(), throw assertion in DBC mode
+  // if this occurs
+  Check(T > bounds.back() * std::numeric_limits<double>::min());
 
   // Initialize the loop:
   double scaled_frequency = bounds[0] / T;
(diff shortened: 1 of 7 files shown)


src/cdi/CDI.cc
```diff
@@ -267,8 +267,10 @@ void CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
   planck.resize(groups, 0.0);
 
   // nu/T must be < numeric_limits<double>::max().  So, if T < nu*min(), then
-  // return early with planck == 0.0. This avoids a possible divide by zero.
-  if (T <= bounds[0] * std::numeric_limits<double>::min())
+  // return early with planck == 0.0. This avoids a possible divide by zero but
+  // it doesn't catch the case where bounds[0] == 0.0 and T is a denormal so
+  // make it the more restrictive T <= numeric_limits<double>::min()
+  if (T <= std::numeric_limits<double>::min())
```
KineticTheory: I think this change breaks the previous logic.  I believe the issue was the case where `bounds[0]` is large so that `bounds[0]/T` would overflow when T>min.  In this scenario, your change makes the check less restrictive and enables the possibility of overflow.

However, your change brings up the case where bounds[0] is small and `bounds[0]*min` underflows.  Maybe we need are more complex set of logic.  Something like:

```c++
if(bounds[0]>1.0) { 
  if (T <= bounds[0] * std::numeric_limits<double>::min())
    return;
} else {
  if (T <= std::numeric_limits<double>::min()) // can this ever occur?
    return;
}
```

src/cdi/CDI.cc
```diff
@@ -281,16 +283,12 @@ void CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
     double const last_planck = planck_value;
 
     // New values:
-    if (T <= bounds[group + 1] * std::numeric_limits<double>::min())
-      planck[group] = 0.0;
```
KineticTheory: Again -- I think this use case is still valid if `bounds[group+1]>1.0`

src/cdi/CDI.cc
```diff
@@ -316,8 +314,11 @@ CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
   std::vector<double> planck(groups, 0.0);
 
   // nu/T must be < numeric_limits<double>::max().  So, if T < nu*min(), then
-  // return early with planck == 0.0. This avoids a possible divide by zero.
-  if (T <= bounds[0] * std::numeric_limits<double>::min())
+  // return early with planck == 0.0. This avoids a possible divide by zero but
```
KineticTheory: ditto

src/cdi/CDI.cc
```diff
@@ -330,16 +331,12 @@ CDI::integrate_Planckian_Spectrum(std::vector<double> const &bounds,
     double const last_planck = planck_value;
 
     // New values:
-    if (T <= bounds[group + 1] * std::numeric_limits<double>::min())
-      planck[group] = 0.0;
```
KineticTheory: ditto

src/cdi/CDI.cc
```diff
@@ -365,8 +362,10 @@ void CDI::integrate_Rosseland_Spectrum(std::vector<double> const &bounds,
   rosseland.resize(groups, 0.0);
 
   // nu/T must be < numeric_limits<double>::max().  So, if T < nu*min(), then
-  // return early with rosseland == 0.0. This avoids a possible divide by zero.
-  if (T <= bounds[0] * std::numeric_limits<double>::min())
+  // return early with rosseland == 0.0. This avoids a possible divide by zero
```
KineticTheory: ditto

src/cdi/CDI.cc
```diff
@@ -385,18 +384,14 @@ void CDI::integrate_Rosseland_Spectrum(std::vector<double> const &bounds,
...
-      scaled_frequency = bounds[group + 1] / T;
-      Check(scaled_frequency > last_scaled_frequency);
-      exp_scaled_frequency = std::exp(-scaled_frequency);
-      integrate_planck_rosseland(scaled_frequency, exp_scaled_frequency,
-                                 planck_value, rosseland_value);
-
-      // Record the definite integral between frequencies.
-      rosseland[group] = rosseland_value - last_rosseland;
-    }
+    scaled_frequency = bounds[group + 1] / T;
```
KineticTheory: ditto

src/cdi/CDI.cc
```diff
@@ -426,8 +421,10 @@ void CDI::integrate_Rosseland_Planckian_Spectrum(
 
   // nu/T must be < numeric_limits<double>::max().  So, if T < nu*min(), then
   // return early with planck ==0.0 and rosseland == 0.0. This avoids a possible
-  // divide by zero
-  if (T <= bounds[0] * std::numeric_limits<double>::min())
```
KineticTheory: ditto

src/cdi/CDI.cc
```diff
@@ -447,21 +444,15 @@ void CDI::integrate_Rosseland_Planckian_Spectrum(
     double const last_rosseland = rosseland_value;
 
     // New values:
-    if (T <= bounds[group + 1] * std::numeric_limits<double>::min()) {
-      planck[group] = 0.0;
```
KineticTheory: ditto

src/ds++/Constexpr_Functions.hh
```diff
@@ -0,0 +1,48 @@
...
+             : std::numeric_limits<double>::quiet_NaN();
+}
+
+} // namespace rtt_dsxx
+
+#endif // rtt_dsxx_Constexpr_Functions_hh
+
+//---------------------------------------------------------------------------//
+// end of ds++/Constexpr_Functions.hh
+//---------------------------------------------------------------------------//
```
KineticTheory: I will send you a patch file that fixes build issues for MSVC.

PR #651, opened 2019-08-02
Implement a rudimentary charged-particle data interface in CDI

keadyk:
### Purpose of Pull Request

* Get a working analytic data interface in place for CP, largely as a demo / first cut to define the interface

### Description of changes

* Mimic inheritance structure for other types of CDI data
* Provide a simple fit for energy loss rate for one type of charged-particle interaction
* Add an approximate evaluation of the error function (Erf), to be used in a future analytic CP data model (coming in a future PR)

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass


src/cdi_analytic/Analytic_CP_Eloss.hh
```diff
@@ -0,0 +1,126 @@
...
+};
+
+//---------------------------------------------------------------------------//
+// INLINE FUNCTIONS
+//---------------------------------------------------------------------------//
+/*!
+ * \brief Return the energy policy descriptor (gray for
+ * Analytic_CP_Eloss).
+ */
+Analytic_CP_Eloss::std_string
```
KineticTheory: I find this return type to be overly verbose.  Maybe just use `std::string`.  Do whatever you think is best for this one.

src/cdi_analytic/Analytic_Models.cc
```diff
@@ -429,6 +429,33 @@ Constant_Analytic_EICoupling_Model::get_parameters() const {
...
+ *
+ * \param T material temperature field in keV
+ * \param rho material density in g/cm^3
+ * \param v incident particle speed in cm/shk
+ * \return eloss (vector of time coefficients) in shk^-1
+ * 
+ */
+double Analytic_KP_Alpha_Eloss_Model::calculate_eloss(double T, double rho,
+                                                      double v) const {
+  double range = 0.03 * T * (1.0 - 0.24 * std::log(1.0 + T)) *
```
KineticTheory: * Add `Require()` entries? especially to rho, T > 0.
* Why 0.37, 0.03, 0.24, etc?  Should these be adjustable parameters?

src/special_functions/Erf.hh
```diff
@@ -0,0 +1,26 @@
...
+//---------------------------------------------------------------------------//
+
+#ifndef sf_Erf_hh
+#define sf_Erf_hh
+
+#include "ds++/config.h"
+
+namespace rtt_sf {
+//! Calculate approximate error function at x
+DLL_PUBLIC_special_functions double Erf(double x);
```
KineticTheory: You can remove all of the DLL_PUBLIC* decorations.

src/special_functions/Erf.hh
```diff
@@ -0,0 +1,26 @@
...
+
+} // end namespace rtt_sf
+
+#endif // sf_F12_hh
+
+//---------------------------------------------------------------------------//
+// end of sf/F12.hh
```
KineticTheory: Fix footer info.

src/cdi/CPCommon.hh
```diff
@@ -0,0 +1,46 @@
...
+};
+
+enum CParticleType {
+  UNKNOWN_PART = 0, /*!< unknown particle :-O */
+  ELECTRON = 1,     /*!< rest are pretty self-explanatory... */
+  HYDROGEN = 2,
+  DEUTERIUM = 3,
+  TRITIUM = 4,
+  ALPHA = 5
+};
```
brryan: Would it be reasonable to use integer ZAIDs (with electrons = -1) in the code instead of this enum? This might lead to a bit less typing when adding particle species, and would match the usage of ZAIDs in CPT data, although things like particle mass will still have to be hardcoded behind branching statements.

KineticTheory: This `enum` list might become very large otherwise!  While an `int` might suffice for now, we should also consider a heavier custom class or struct.  Doing so would give us a little more flexibility in defining special types like electrons.  The custom class could have ctor and getters that take/give an integer ZAID value so it could be used like an `int`

david8dixon: Is there a reason why you are using branching as as opposed to tag dispatching? You could have tags with fixed particles properties built in:

```c++
struct Alpha {
  static constexpr auto mass(){ return 6.644657230E-27 * kilo(grams); }
  static constexpr auto charge(){ return 3.2043532416E-19 * coulomb; }
};
```
Then rather than check a condition everytime you fall through a function, you just dispatch to the appropriate function based on the particle tag. For example, this:

```c++
auto foo(Particle p){
  // setup
  if(particle.type() == CParticleType::ELECTRON){
    // implementation electron
  } else if (particle.type() == CParticleType::ALPHA) {
  // implementation for alpha
 }...
}
```
can then become:

```c++
void do_something_impl(Particle p, Electron){
  // do some stuff
}
void do_something_impl(Particle p, Alpha){
  // do some stuff
}
void do_something(Particle p){
  do_something_impl(p, Particle::Type{});
}

auto foo(Particle p){
  // setup
  do_something(p);
}
```
The benefit of doing it this way is that you do not branch at runtime because the compiler will instantiate a function for the different particles types and call it accordingly. Furthermore, this approach follows the open/close principle in that you can accomodate new particles by simply adding a new function to the overload set as opposed to modify an existing function. Happy to discuss this in person if it is unfamiliar.

warsa: Maybe even simpler, see https://cppinsights.io/s/38e452e1

warsa: Why not use an enum class, where the names are local to the enum and their values do not implicitly convert to other types, and without assignment?
enum class CPType
{
  CPElectron,
  CPAlpha,
  CPNone
}

keadyk: Thank you all for the input! I understand what you're saying with regard to the design of the class. I admittedly copy-pasted this code from an existing CDI example to get something I could build against. I will address the simpler comments (formatting, etc.) and leave the WIP in place until we hash this out :)

src/cdi/CPEloss.hh
```diff
@@ -0,0 +1,180 @@
...
+   * object derived from CPEloss is destroyed.
+   */
+  virtual ~CPEloss(){/*empty*/};
+
+  // --------- //
+  // Accessors //
+  // --------- //
+
+  /*!
+     * \brief Eloss accessor that returns a single eloss that
```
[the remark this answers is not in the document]
keadyk: Done!

keadyk:
First off, don't be alarmed, anyone, I force-pushed to clean up several commits into one. 

@brryan My questions for you-- Does this look more like you were thinking (wrt the ZAIDs)? Does the base class have everything we'll need to implement a tabular CPEloss model? Are you okay with my current units for Eloss? I went with per unit time (de/dt as opposed to de/dx), just because that's how we typically think of transport steps... but I am completely open to whatever works best for you.

@KineticTheory I've deviated from the traditional CDI design somewhat in this iteration. We typically size the CDI SP arrays when we construct a CDI object, with space for all possible combos of reactions. That didn't seem feasible here, so I construct an empty energy loss SP array, then push_back into it as models are added. There's a corresponding map from index -> projectile/target pair that is also updated.  Naturally, avoiding push_back would be preferred, but I didn't see a clear way to do this. I'd also like to get away from shared pointers in CDI (in favor of unique pointer transfers), but that's probably a task for another day :)

src/cdi/CPCommon.hh
```diff
@@ -0,0 +1,48 @@
...
+enum class CPModel {
+  ELOSS = 0,        /*!< energy loss model */
+  LA_SCATTERING = 1 /*!< large-angle scattering model */
+};
+
+enum class CPModelType {
+  UNASSIGNED_ETYPE =
+      0, /*!< unassigned type; used as a placeholder before deciding type */
+  ANALYTIC_ETYPE = 1, /*!< an Analytic eloss model */
+  TABULAR_ETYPE = 2   /*!< an Analytic eloss model */
```
brryan: /*!< a Tabular eloss model */

src/cdi/CPCommon.hh
```diff
@@ -0,0 +1,48 @@
...
+// Basic data required to describe a charged particle
+class CParticle {
+private:
+  int32_t zaid;
+  double mass;
+
+public:
+  int32_t get_zaid() const { return zaid; }
+  double get_mass() const { return mass; }
+};
```
brryan: Just a thought: what about adding a vector of references to CPEloss classes (and later classes containing large angle scattering physics) here, so that with this class you get both the charged particle properties as well as the reactions it can undergo? Might make bookkeeping more convenient later.

src/cdi/CPEloss.hh
```diff
@@ -0,0 +1,141 @@
...
+   *
+   * \return A single Eloss rate coefficient (shk^-1).
+   */
+  virtual double getEloss(const double targetTemperature,
+                          const double targetDensity,
+                          const double partSpeed) const = 0;
+  /*!
+   * \brief Query whether the data is in tables or functional form.
+   */
+  virtual bool data_in_tabular_form() const = 0;
```
brryan: Is there a use case for this function?

src/cdi/CPEloss.hh
```diff
@@ -0,0 +1,141 @@
...
+
+  /*! 
+   * \brief Returns the size of the density grid.
+   */
+  virtual size_t getNumDensities() const = 0;
+
+  /*!
+   * \brief Returns the size of the energy grid.
+   */
+  virtual size_t getNumEnergies() const = 0;
```
brryan: Can these functions be declared only in the tabular Eloss inherited class?

src/cdi_analytic/Analytic_CP_Eloss.hh
```diff
@@ -0,0 +1,120 @@
...
+  sf_double getEnergyGrid() const { return sf_double(0); }
+
+  //! Get the size of the temperature grid (size 0).
+  size_t getNumTemperatures() const { return 0; }
+
+  //! Get the size of the density grid (size 0).
+  size_t getNumDensities() const { return 0; }
+
+  //! Get the size of the density grid (size 0).
+  size_t getNumEnergies() const { return 0; }
```
brryan: Similar to above comment, do we need these functions in the analytic Eloss class?

src/special_functions/Erf.cc
```diff
@@ -0,0 +1,58 @@
...
+ * \author Kendra Long
+ * \date   Wed Jul 31 08:18:59 2019
+ * \brief  Implementation of Erf approximation
+ * \note   Copyright (C) 2019 Triad National Security, LLC.
+ *         All rights reserved. */
+//---------------------------------------------------------------------------//
+
+#include "Erf.hh"
+#include "units/MathConstants.hh"
+#include <cmath>
```
brryan: I believe `<cmath>` defines 'erf'; can we just use this? Maybe there are performance concerns.

src/cdi/CPEloss.hh
keadyk: Probably not? Not in our code, at least. I hemmed and hawed about several of the pure virtual functions, but ultimately left them in place because I'm not 100% how other codes might use these classes :)

src/cdi/CPCommon.hh
keadyk: Hmm, this is an interesting idea! I guess in my original plan, the CParticle class was intended to provide the necessary input to build a CDI Eloss -- I hadn't thought about using it (or something like it) as a container. My general plan was to implement the Eloss container part of the equation in the calling code, but I'm certainly not tied to that idea if you think it makes more sense

(Also, it just occurred to me that I haven't even provided a way to set the zaid/mass, so obviously this design isn't terribly useful as is. I guess this is what happens when the only actual Eloss model you've implemented needs no particle parameters.)

brryan: My thinking in bringing this up was just to avoid lookups-by-ZAID during compute loops; initializing a CParticle for each physical particle, with references to physics classes ready to go, that is then referenced would be one way around that I think. But there's a bit of a circular dependency issue here, because you kind of need CParticle to initialize CDI Eloss, like you say... unfortunately circular dependencies that don't compile is my natural C++ coding style. Maybe a separate CParticlePhysics class that holds references to both a CParticle and CDI Elosses would be best? Whatever you prefer layout-wise is fine with me, I just think it would be nice to only find and store one reference by zaid when creating e.g. a computational particle (you may have had this in mind all along).

keadyk: I agree, I think that's the way to go for the actual transport step--  I'm just not sure where the container of particle info / eloss references should live. Right now I'm assuming this container is "above" draco in the calling code, but I'm open to other design options :)

brryan: Oh good point, that makes way more sense, leaving those details out of Draco -- please ignore what I said in this thread.

keadyk: Heh, no worries. In any case, the CParticle class will definitely need to change so you can... you know... actually set the class members :D

brryan:
LGTM

PR #685, opened 2019-10-22
Add extra checks for gcc-Debug builds

KineticTheory:
### Background

* During C++ training by Jason Turner (C++ class 2019-Oct), these flags (along with others) for gcc were recommended.

### Description of changes

+ Adding `-Wconversion -Wdouble-promotion -Wshadow -Wformat=2` and fix issues related to these warnings.
+ Remove `DLL_PUBLIC_c4` from most of `c4`. A few global static variables still need this decoration.
+ Notes:
  + gcc flags `size_t` -> `double` conversions.  So, there are more additions of `static_cast<double>(...)` calls than I would have expected.
  + Most of the issues with _shadowing_ occur in constructors and member functions that mask class variables.  In these cases, I simply appended the function arguments with `_in`.
  + I probably should have kept the `DLL_PUBLIC` removal in a separate PR, but these changes got intertwined in my sandbox. The delta is straight forward and shouldn't add too much burden to the reviewer.

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass


config/unix-g++.cmake
```diff
@@ -90,11 +90,13 @@ if( NOT CXX_FLAGS_INITIALIZED )
   if( CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 7.0 )
     string( APPEND CMAKE_C_FLAGS    " -Wno-expansion-to-defined -Wnarrowing" )
   endif()
-  set( CMAKE_C_FLAGS_DEBUG          "-g -gdwarf-3 -fno-inline -fno-eliminate-unused-debug-types -O0 -Wextra -Wundef -Wunreachable-code -DDEBUG")
+  set( CMAKE_C_FLAGS_DEBUG          "-g -gdwarf-3 -fno-inline -fno-eliminate-unused-debug-types -O0 -Wextra -Wundef -Wunreachable-code -Wconversion -Wdouble-promotion -Wshadow -Wformat=2 -DDEBUG")
```
KineticTheory: Move these to `src/CMakeLists.txt` for now to avoid impacting other codes that pull their flags from this file.

src/c4/test/tstProcessor_Group.cc
```diff
@@ -52,7 +53,7 @@ void tstProcessor_Group(rtt_dsxx::UnitTest &ut) {
     vector<double> myvec;
     size_t const vlen(5);
     for (size_t i = 0; i < vlen; ++i)
-      myvec.push_back(pid * 1000.0 + i);
+      myvec.push_back(static_cast<double>(pid * 1000 + i));
```
KineticTheory: This is ugly.  Learn how to enable `-Wconversion` w/o warning on `int --> double` promotion when multiplying (if this is even possible).

config/unix-g++.cmake
KineticTheory: fixed.

src/c4/test/tstProcessor_Group.cc
KineticTheory: The `uint64_t` type is exact, but when converted to a double it might loose its _exactness_.

KineticTheory:
@along This is ready for final review.  All tests pass.

src/ds++/test/tstSP.cc
```diff
@@ -674,7 +671,7 @@ void type_X_test(rtt_dsxx::UnitTest &ut) {
       CHECK_N_OBJECTS(3, 3, 0, 0);
 
       // assign SPfoo2 to itself
-      spfoo2 = spfoo2;
+      // spfoo2 = spfoo2; // modern compilers flag this at compile time.
```
alexrlongne: Does this need to stay in the code?

src/ds++/test/tstDracoMath.cc
```diff
@@ -17,7 +17,7 @@
 //---------------------------------------------------------------------------//
 
 void tstabs(rtt_dsxx::UnitTest &ut) {
-  using rtt_dsxx::abs;
+  using std::abs;
```
keadyk: I feel like the purpose of this is to test the overloads in ds++/DracoMath:

`// Specialization for standard types - There is no standard abs function for
// float -- one reason why we define this template!
template <> inline double abs(double a) { return std::fabs(a); }
template <> inline int abs(int a) { return std::abs(a); }
template <> inline long abs(long a) { return std::labs(a); }`

So this might be changing the intent / behavior of the test (even though the test still passes). It looks like it does decrease the coverage of the DracoMath header!

src/rng/LC_Subrandom_Generator.cc
```diff
@@ -14,11 +14,12 @@ namespace rtt_rng {
...
- * \param count_ Dimension of the vector of random values returned by this
+ * \param count_in Dimension of the vector of random values returned by this
  *          object.
  */
-LC_Subrandom_Generator::LC_Subrandom_Generator(unsigned const count_)
-    : Subrandom_Generator(count_), generator_(gsl_rng_alloc(gsl_rng_mt19937)) {
+LC_Subrandom_Generator::LC_Subrandom_Generator(unsigned const count_in)
+    : Subrandom_Generator(count_in),
+      generator_(gsl_rng_alloc(gsl_rng_mt19937)) {
   // generator_ = gsl_rng_alloc(gsl_rng_mt19937);
```
keadyk: This line can presumably be deleted :)

src/ds++/test/tstSP.cc
KineticTheory: I want to leave this in the code for now.  Originally, this checked that our home-grown smart-pointer class did the right thing when self-assigned (i.e. no increment of object counter).

src/rng/LC_Subrandom_Generator.cc
KineticTheory: good catch.  I'll remove this.

config/unix-g++.cmake
```diff
@@ -95,7 +95,9 @@ if( NOT CXX_FLAGS_INITIALIZED )
   # -Wfloat-equal
   # -Werror
   # -Wconversion
+  # -Wdouble-promotion
```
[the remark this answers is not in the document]
KineticTheory: I'll do this.

src/ds++/test/tstDracoMath.cc
KineticTheory: I've removed `rtt_dsxx::abs` altogether.  I don't see why it is needed and `std::abs` seems to do everything we need.

KineticTheory:
@warsa This PR removes `rtt_dsxx::abs`.  Does this cause any issues for you?  I couldn't find any reason to use `rtt_dsxx::abs` over [`std::abs`](https://en.cppreference.com/w/cpp/numeric/math/fabs) and the coding for `rtt_dsxx::abs` had issues that resulted in build warnings.

warsa:
@KineticTheory @jhchang-lanl 
I don't think so, but I'd say don't worry about it and I'll build/test and look for and accommodate any changes, if necessary, when this PR is merged.

PR #697, opened 2019-11-06
CDI tabulated stopping power reader

brryan:
### Background

* Some expressions for charged particle stopping powers are expensive to evaluate, and are better tabulated. Here we add CDI support for one stopping power datatable format by inheriting from the `CPEloss` class.

### Purpose of Pull Request

* Add `Tabular_CP_Eloss` class, which inherits from `CPEloss` in a new `cdi_CPEloss` module and can read and parse stopping power data tables

### Description of changes

* Change `CPEloss` so that it accepts references to `CParticle`s for target and projectile input, rather than ZAIDs (this is one choice to address the issue from before of whether `CParticle`s contain interaction physics or vice verse)
* Added `DracoArray` container for multidimensional access to a single `std::vector`
* Removed enum distinguishing between stopping powers and large angle scattering, since large angle scatterings may be better left to a separate class from `CPEloss`
* Added `CPModelAngleCutoff` enum, to indicate which angle cutoff is used by a stopping power model (which would inform how large angle scattering physics is implemented)
* Added algorithm for 3D linear interpolation, `linear_interpolate_3` on a uniform lattice to `DracoMath.hh` (needed for interpolating 3D stopping power data)

### TODO

* @keadyk I made some fairly significant changes to `CPEloss` so please do object if you have issues with what I've done
* @kgbudge KT and I dropped by your office earlier about multidimensional arrays in Draco; if you have opinions about the `DracoArray` included here I'm happy to discuss -- this is just some code I had lying around already
* Remove virtual member functions from `CPEloss`?
* Add tests
* Formatting

### Status

* Reference:
* Checks
  * [ ] Travis CI checks pass


src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+ * LANL format for stopping powers. 
+ *
+ * \param filename_in path to eloss file
+ * \param target_zaid_in target particle zaid
+ * \param projectile_zaid_in transporting particle zaid
+ */
+Tabular_CP_Eloss::Tabular_CP_Eloss(std::string filename_in,
+                                   rtt_cdi::CParticle &target_in,
+                                   rtt_cdi::CParticle &projectile_in)
+    : filename(filename_in), rtt_cdi::CPEloss(target_in, projectile_in) {
```
keadyk: I think the base class constructor call should always come first in the initializer list, because it always gets called first in practice...

src/cdi_analytic/Analytic_CP_Eloss.cc
```diff
@@ -30,15 +30,15 @@ namespace rtt_cdi_analytic {
  * \param projectile_zaid_in int32_t transporting particle zaid
  */
 Analytic_CP_Eloss::Analytic_CP_Eloss(SP_Analytic_Model model_in,
-                                     int32_t target_zaid_in,
-                                     int32_t projectile_zaid_in)
-    : analytic_model(model_in), projectile_zaid(projectile_zaid_in),
-      target_zaid(target_zaid_in) {
+                                     rtt_cdi::CParticle &target_in,
+                                     rtt_cdi::CParticle &projectile_in)
+    : analytic_model(model_in), rtt_cdi::CPEloss(target_in, projectile_in) {
```
keadyk: Same here, I think the base class constructor call needs to be called before analytic_model is set :)

src/cdi_CPEloss/Tabular_CP_Eloss.cc
brryan: Already fixed this and the next related issue on my local branch (they generate compiler warnings that I noticed too late), but hadn't pushed yet to avoid spamming everyone with "new commit" emails. Good catch though!

src/cdi/CPEloss.hh
```diff
@@ -34,6 +34,8 @@ namespace rtt_cdi {
 
 class CPEloss {
 public:
+  CPEloss(CParticle &target_in, CParticle &projectile_in)
+      : projectile(projectile_in), target(target_in) {}
```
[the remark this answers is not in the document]
brryan: I wondered about this too, and I honestly don't know whether references are the right thing to do here

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+  if (!file.is_open()) {
+    std::cout << "Eloss file " << filename << " could not be opened!"
+              << std::endl;
+    exit(-1);
+  }
+
+  std::string line;
+  std::vector<std::string> line_entries;
+  int nlines;
+  int max_entries =
```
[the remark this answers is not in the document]
brryan: Oops yes you're right, this can be a constexpr

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+        // This is not the requested target ion
+        for (int n = 0; n < nlines; n++) {
+          read_line();
+        }
+      }
+    }
+  }
+  file.close();
+
+  if (!target_found) {
```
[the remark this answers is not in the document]
brryan: Yes I should do this, I was being lazy and non-Draco-y

src/ds++/DracoArray.hh
```diff
@@ -0,0 +1,135 @@
...
+  int get_n3() const {
+    Require(dim > 2);
+    return n3;
+  }
+  std::vector<int> get_size() {
+    std::vector<int> size_all{n1, n2, n3};
+    std::vector<int> size;
+    for (int n = 0; n < dim; n++) {
+      size.push_back(size_all[n]);
+      return size;
```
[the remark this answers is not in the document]
brryan: Good catch, this will teach me to create WIPs before writing up any unit testing...

KineticTheory:
@brryan To fix the failing CI testing try this on one of the ccs-net machines

```console
# setup environment
$ module list

Currently Loaded Modules:
  1) user_contrib      7) htop/2.0.2      13) totalview/2019.0.4            19) netlib-lapack/3.8.0  25) csk/0.4.1
  2) ack/2.22          8) mscgen/0.20     14) gcc/8.1.0                     20) qt/5.12.5            26) ndi/2.1.3
  3) dia/0.97.3        9) numdiff/5.9.0   15) cmake/3.14.3                  21) openmpi/3.1.0
  4) doxygen/1.8.14   10) python/3.6.5    16) eospac/6.3.1_r20161202150449  22) parmetis/4.0.3
  5) git/2.23.0       11) random123/1.09  17) gsl/2.4                       23) superlu-dist/5.2.2
  6) graphviz/2.40.1  12) tk/8.6.8        18) metis/5.1.0     

# regular debug build

$ mkdir -p /scratch/brryan/gcc-mpid/d && cd $_
$ C_FLAGS=-Werror CXX_FLAGS=-Werror cmake ~/draco
$ make -j 20

# scalar debug build

$ mkdir -p /scratch/brryan/gcc-scalard/d && cd $_
$ C_FLAGS=-Werror CXX_FLAGS=-Werror cmake -DDRACO_C4=SCALAR ~/draco
$ make -j 20

# autodoc

$ cd /scratch/brryan/gcc-mpid/d
$ make autodoc &> autodoc.log
$ vi autodoc.log # <-- look for the regex 'error'
```

src/cdi/CPCommon.hh
```diff
@@ -20,9 +20,10 @@ namespace constants {
 unsigned int const num_CPModels(3);
 } // namespace constants
 
-enum class CPModel {
-  ELOSS = 0,        /*!< energy loss model */
-  LA_SCATTERING = 1 /*!< large-angle scattering model */
+enum class CPModelAngleCutoff {
+  NONE = 0, /*! no angle cutoff */
```
KineticTheory: Use trailing enumeration doxygen comments.  E.g.: `/*!< no angle cutoff */`

src/cdi/CPEloss.hh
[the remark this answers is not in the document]
KineticTheory: The question comes down to whether `CParticle` is trivially constructable.  If so, then pass by copy, otherwise I think you want to use `const &`.

src/cdi/CPEloss.hh
```diff
@@ -129,7 +132,40 @@ public:
 	 * \brief Returns the general Eloss model type (Analytic),
 	 * defined in the enum at the top of this file.
 	 */
-  virtual rtt_cdi::CPModelType getModelType() const = 0;
+  //virtual rtt_cdi::CPModelType getModelType() const = 0;
+
+  /*!
+   * \brief Query to determine the transporting particle type.
+   */
+  CParticle getProjectile() const { return projectile; }
```
KineticTheory: What do you think about making projectile a public data member?  This would allow us to eliminate these _getters_ from the class implementation.  Also consider if you want to return a reference to the _projectile_ or a copy.

src/cdi_CPEloss/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+  add_subdirectory( test )
+endif()
+
+# ---------------------------------------------------------------------------- #
+# Autodoc
+# ---------------------------------------------------------------------------- #
+
+process_autodoc_pages()
+
+# ---------------------------------------------------------------------------- #
```
KineticTheory: This file is fine for now.  David is helping me re-imagine our `CMakeLists.txt` files to make better use of modern CMake features.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+ * \date   2019 Nov 4                                             
+ * \brief  Tabular_CP_Eloss member definitions.                                 
+ * \note   Copyright (C) 2016-2019 Triad National Security, LLC.                 
+ *         All rights reserved. */
+//---------------------------------------------------------------------------//
+
+#include "Tabular_CP_Eloss.hh"
+
+using std::stod;
+using std::stoi;
```
KineticTheory: Technically this is safe because the namespace injection is local to the current translation unit.  However, I would prefer to see these `using` statements inside function scopes where they are needed.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * This constructor builds an eloss model defined by the
+ * rtt_cdi_cpeloss::Tabular_Eloss_Model derived class argument.
+ *
+ * The path to an eloss datafile is passed to the constructor,
+ * which opens and parses the file. The file format is the usual
+ * LANL format for stopping powers. 
+ *
+ * \param filename_in path to eloss file
+ * \param target_zaid_in target particle zaid
```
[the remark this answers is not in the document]
KineticTheory: Also, please use `\param[in]` to indicate the design that these variables are not returned from the function call.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+  model_type = rtt_cdi::CPModelType::TABULAR_ETYPE;
+
+  file.open(filename);
+  if (!file.is_open()) {
+    std::cout << "Eloss file " << filename << " could not be opened!"
+              << std::endl;
+    exit(-1);
+  }
+
+  std::string line;
```
KineticTheory: Please delay declaration until these variables are actually needed.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+      read_line(); // Bin spacing for energy, density, temperature (log)
+  d_log_energy = 1. / stod(line_entries[0]);
+  d_log_density = 1. / stod(line_entries[1]);
+  d_log_temperature = 1. / stod(line_entries[2]);
+
+  // Get first energy support point
+  nlines = std::ceil((double)n_energy / max_entries);
+  line_entries = read_line();
+  min_log_energy = stod(line_entries[0]);
+  for (int n = 0; n < nlines - 1; n++) {
```
KineticTheory: Probably change `n` to an unsigned type.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  std::string line;
+  std::vector<std::string> line_entries;
+  int nlines;
+  int max_entries =
+      6; // This is a statement about the file format, maximum of six entries for row.
+
+  read_line(); // ZAID
+  int32_t projectile_zaid_file = stoi(line_entries[0]);
```
keadyk: I think you meant to set line_entries = read_line() on line 55 -- otherwise 56 will segfault! :O

src/cdi/CPEloss.hh
brryan: My thinking was to ensure the target/projectile members cannot be modified after construction, is a getter function an OK way to indicate this? Also, from the discussion about passing target/projectile by copy rather than reference, should it also be a copy that gets returned?

src/cdi_CPEloss/Tabular_CP_Eloss.cc
brryan: 🤦‍♂️

brryan:
> @brryan To fix the failing CI testing try this on one of the ccs-net machines
> 
> ```
> # setup environment
> $ module list
> 
> Currently Loaded Modules:
>   1) user_contrib      7) htop/2.0.2      13) totalview/2019.0.4            19) netlib-lapack/3.8.0  25) csk/0.4.1
>   2) ack/2.22          8) mscgen/0.20     14) gcc/8.1.0                     20) qt/5.12.5            26) ndi/2.1.3
>   3) dia/0.97.3        9) numdiff/5.9.0   15) cmake/3.14.3                  21) openmpi/3.1.0
>   4) doxygen/1.8.14   10) python/3.6.5    16) eospac/6.3.1_r20161202150449  22) parmetis/4.0.3
>   5) git/2.23.0       11) random123/1.09  17) gsl/2.4                       23) superlu-dist/5.2.2
>   6) graphviz/2.40.1  12) tk/8.6.8        18) metis/5.1.0     
> 
> # regular debug build
> 
> $ mkdir -p /scratch/brryan/gcc-mpid/d && cd $_
> $ C_FLAGS=-Werror CXX_FLAGS=-Werror cmake ~/draco
> $ make -j 20
> 
> # scalar debug build
> 
> $ mkdir -p /scratch/brryan/gcc-scalard/d && cd $_
> $ C_FLAGS=-Werror CXX_FLAGS=-Werror cmake -DDRACO_C4=SCALAR ~/draco
> $ make -j 20
> 
> # autodoc
> 
> $ cd /scratch/brryan/gcc-mpid/d
> $ make autodoc &> autodoc.log
> $ vi autodoc.log # <-- look for the regex 'error:'
> ```

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+  std::ostringstream file_error;
+  file_error << "Error opening DEDX file \"" << filename << "\"";
+  Insist(file.is_open(), file_error.str());
+
+  std::vector<std::string> line_entries;
+  uint32_t nlines;
+  constexpr uint32_t max_entries =
+      6; // This is a statement about the file format, maximum of six entries per row.
+
+  line_entries = read_line(); // ZAID
```
KineticTheory: I would combine lines 46 and 51:

```c++
std::vector<std::string> line_entries = read_line();
```

and the same for `nlines`.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+  int32_t projectile_zaid_file = stoi(line_entries[0]);
+  Require(projectile.get_zaid() == projectile_zaid_file);
+
+  read_line(); // Z, A, mass
+
+  line_entries = read_line(); // Number of bins for energy, density, temperature
+  n_energy = stoi(line_entries[0]);
```
KineticTheory: Consider adding `Check(line_entries.size() >= 3)` before accessing these elements.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+
+  line_entries =
+      read_line(); // Bin spacing for energy, density, temperature (log)
+  d_log_energy = 1. / stod(line_entries[0]);
+  d_log_density = 1. / stod(line_entries[1]);
+  d_log_temperature = 1. / stod(line_entries[2]);
+
+  // Get first energy support point
+  nlines =
+      static_cast<int>(std::ceil(static_cast<double>(n_energy) / max_entries));
```
KineticTheory: I think you can save the FP cycles by using integer division: `nlines = (n_energy+max_entries -1) / max_entries`.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+      ((double)n_energy * n_density * n_temperature) /
+      max_entries)); // The number of lines taken up by stopping power data for one target
+  if (target.get_zaid() == -1) {
+    // Target is free electrons
+    target_found = true;
+    uint32_t nentry = 0;
+    for (uint32_t n = 0; n < nlines; n++) {
+      line_entries = read_line();
+      for (std::string entry : line_entries) {
+        stopping_data_1d[nentry] = stod(entry);
```
KineticTheory: Please double check, but I think you can eliminate line 121 by saying `stopping_data_1d[nentry++] = stod(entry);`

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+  d_log_temperature = 1. / stod(line_entries[2]);
+
+  // Get first energy support point
+  nlines =
+      static_cast<int>(std::ceil(static_cast<double>(n_energy) / max_entries));
+  line_entries = read_line();
+  min_log_energy = stod(line_entries[0]);
+  for (uint32_t n = 0; n < nlines - 1; n++) {
+    read_line();
+  }
```
KineticTheory: This pattern occurs frequently enough that I think you should create a `read_lines(uint32_t nlines)` helper function.  Or, you could update `read_line` to take an optional `nlines` argument.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+        // This is not the requested target ion
+        for (uint32_t n = 0; n < nlines; n++) {
+          read_line();
+        }
+      }
+    }
+  }
+  file.close();
+
+  std::ostringstream dedx_error;
```
KineticTheory: You might want to avoid creating this ostringstream object as I mentioned above.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+        stopping_data(ne, nd, nt) =
+            stopping_data_1d[ne + n_energy * (nd + n_density * nt)];
+      }
+    }
+  }
+
+  // Convert units on table to match those of getEloss:
+  //   energy:      MeV -> cm/shk (using target particle mass)
+  double energy_cgs = exp(min_log_energy) * (1.e6 * 1.6021772e-12);
+  min_log_energy = log(sqrt(2. * energy_cgs / target.get_mass()) * 1.e-8);
```
KineticTheory: Let's talk about this unit conversion.  We might want to use entries from the `units` package to ensure we use the same conversion factors in all of our code.  The `units` package has some known deficiencies that @along and I are discussing, but I think it is better to use it for now.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/*!
+ * \brief Read a line from an eloss datafile and return as a vector of strings.
+ *
+ * Convenience function, especially when reading header data which is not
+ * uniform in the number of entries per line, or the types of those entries.
+ *
+ * \return entries the resulting vector of entries in the datafile line.
+ *
```
KineticTheory: These extra blank lines drive me a little crazy.  Call me overly-pedantic, but can you remove this?

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+std::vector<std::string> Tabular_CP_Eloss::read_line() {
+  std::string line;
+  std::getline(file, line);
+  std::vector<std::string> entries;
+  std::istringstream iss(line);
+  for (std::string s; iss >> s;) {
+    entries.push_back(s);
+  }
+  return entries;
+}
```
KineticTheory: @keadyk Do you have any thoughts about making this function less verbose?

src/cdi_CPEloss/Tabular_CP_Eloss.cc
```diff
@@ -0,0 +1,270 @@
...
+  Require(partSpeed < max_energy);
+
+  if (temperature < min_temperature || temperature > max_temperature ||
+      density < min_density || density > max_density ||
+      partSpeed < min_energy || partSpeed > max_energy) {
+    // Outside of the table
+    return 0.;
+  }
+
+  int pt0_energy = static_cast<int>(
```
KineticTheory: Mark as many of these as possible `const`.

src/cdi_CPEloss/Tabular_CP_Eloss.hh
```diff
@@ -0,0 +1,100 @@
...
+
+class Tabular_CP_Eloss : public rtt_cdi::CPEloss {
+public:
+  typedef std::vector<double> sf_double;
+
+private:
+  std::string filename;
+  std::ifstream file;
+
+  uint32_t n_energy;          // Number of gridpoints in projectile energy
```
KineticTheory: Change comment trailing style from `// comment` to `//!< comment`.

src/cdi_CPEloss/Tabular_CP_Eloss.hh
```diff
@@ -0,0 +1,100 @@
...
+  sf_double energies;         // Vector of energy gridpoints
+  sf_double densities;        // Vector of density gridpoints
+  sf_double temperatures;     // Vector of temperature gridpoints
+  // Note that after unit conversions, *_energy is really *_speed
+
+  // Storage for tabulated data
+  rtt_dsxx::DracoArray<double> stopping_data;
+
+  // Utility for reading a line of an eloss file and as a vector of strings
+  std::vector<std::string> read_line();
```
KineticTheory: Can this be const?

src/cdi_CPEloss/Tabular_CP_Eloss.hh
```diff
@@ -0,0 +1,100 @@
...
+ *                                                                               
+ */
+//===========================================================================//
+
+class Tabular_CP_Eloss : public rtt_cdi::CPEloss {
+public:
+  typedef std::vector<double> sf_double;
+
+private:
+  std::string filename;
```
KineticTheory: Can this be const?

src/cdi_CPEloss/Tabular_CP_Eloss.hh
KineticTheory: Can any be const or constexpr?

src/cdi_CPEloss/Tabular_CP_Eloss.cc
brryan: Added a similar check for the bin spacing line.

src/cdi_CPEloss/Tabular_CP_Eloss.cc
brryan: I didn't think this would work but it does (I checked in a separate test code, and confirmed online that `expr++` returns `expr` before incrementing)

src/cdi_CPEloss/Tabular_CP_Eloss.cc
brryan: Thanks this is way cleaner

src/cdi_CPEloss/Tabular_CP_Eloss.cc
brryan: Added `void Tabular_CP_Eloss::skip_lines(uint32_t nlines)` method.

src/cdi/CPEloss.hh
```diff
@@ -34,6 +34,8 @@ namespace rtt_cdi {
 
 class CPEloss {
 public:
+  CPEloss(CParticle target_in, CParticle projectile_in)
```
keadyk: I've been thinking about this constructor (again :D), and I have a question: in the case of tabular data, is there any reason not to just use the particle masses found in the data tables? I'm trying to figure out if this general constructor format (passing in particle objects that already "know" their own basic properties) makes sense for all the inherited Eloss classes, or if we should modify it somehow. In the case of the tabular data, it SEEMS like all we really need to know are the ZAIDs, and the rest can be gleaned from the table... but is there something I'm missing here?

src/cdi_CPEloss/Tabular_CP_Eloss.cc
keadyk: Hmm, my initial thought was:

http://www.cplusplus.com/reference/cstring/strtok/

but this is an old c-style string function, so I don't think it makes sense to hack it in here. Honestly, I'm not sure there's a more elegant solution to tokenizing std::strings!

KineticTheory: In `ds++/DracoStrings.hh`, I have a function `tokenize` that converts a string into a vector of words, but it is more code because it allows any character to be used as a delimiter (not just a space).

` std::vector<std::string> entries = rtt_dsxx::tokenize(line);`

keadyk: Oh right!! I should've remembered that, because I've used it before :)

src/cdi/CPEloss.hh
brryan: The tabular data format read by this class does provide target and projectiles masses (although not the mass of the electron target!) that we could read in. I guess I feel safer though keeping constants like particle masses centralized somewhere in the code, so the code doesn't one day find itself in a situation where a particle has a different mass depending on which line of code it's on. If this constructor has problems though I'm happy to discuss. Another solution could be to have a more global namespace-y depot that can be queried for things like particle masses (and then you would only have to pass ZAIDs to Eloss classes).

keadyk: Cool, I'm fine with your reasoning (and leaving the constructor as-is, at least until someone else complains about it :) ).

src/cdi_CPEloss/Tabular_CP_Eloss.cc
brryan: Looking at `units`, I'm not sure that module supports what we want to do here: converting from a particle energy to a particle speed. I might be able to convert particle energy in eV to Kelvin and then to ergs with Boltzmann, but that feels unsatisfactory.

src/cdi_CPEloss/Tabular_CP_Eloss.hh
brryan: Comment style changed, but as far as I can tell none of these member variables can be `const`/`constexpr` because they are assigned during the constructor from values in the data file.

KineticTheory:
@brryan If this is ready for final review and merge, please remove the `WIP`.

KineticTheory:
@brryan Can `src/cdi_CPEloss/test/001-H-001` be reduced.  This seems like a really big file.  If it is only used for testing I think we need a light weight proxy (fake data).  If this is actual data from another library, we should must be careful about releasing as open-source.  Another option is to just point to existing files on our systems?

brryan:
> @brryan Can `src/cdi_CPEloss/test/001-H-001` be reduced. This seems like a really big file. If it is only used for testing I think we need a light weight proxy (fake data). If this is actual data from another library, we should must be careful about releasing as open-source. Another option is to just point to existing files on our systems?

@KineticTheory I replaced `001-H-001` with a 5.5 KB data file I made myself. I also rebased to fix the merge conflict. Do we want to replace `DracoArray` with `mdspan`?

src/cdi_CPEloss/Tabular_CP_Eloss.hh
```diff
@@ -0,0 +1,107 @@
...
+  double max_density;         //!< Maximum target density
+  double min_temperature;     //!< Minimum target temperature
+  double max_temperature;     //!< Maximum target temperature
+  sf_double energies;         //!< Vector of energy gridpoints
+  sf_double densities;        //!< Vector of density gridpoints
+  sf_double temperatures;     //!< Vector of temperature gridpoints
+  // Note that after unit conversions, *_energy is really *_speed
+
+  // Storage for tabulated data
+  std::experimental::basic_mdspan<double, std::experimental::extents<std::experimental::dynamic_extent, std::experimental::dynamic_extent, std::experimental::dynamic_extent>, std::experimental::layout_left> stopping_data;
```
brryan: @KineticTheory I'm unhappy with this but I'm not sure how to get rid of the `std::experimental`s without adding to the namespace for all source files that include this header.

src/cdi_CPEloss/Tabular_CP_Eloss.hh
```diff
@@ -0,0 +1,113 @@
...
+ *        This class implements the interface found in cdi/CPEloss.hh for        
+ *        the case where CP energy loss data is in tabular form, stored 
+ *        in a file.                
+ *                                                                               
+ */
+//===========================================================================//
+
+class Tabular_CP_Eloss : public rtt_cdi::CPEloss {
+public:
+  typedef std::vector<double> sf_double;
```
alexrlongne: Can you add a couple of typedefs here maybe:
`typedef std::experimental::dynamic_extent dyanmic_extent;`
`typedef std::experimental::layout_left layout_left;`
This should make line 70 a little more clear.

alexrlongne:
@KineticTheory Could you take a look at the appveyer failure? There's a warning in there about unreachable code in a file in `c4` but Ben hasn't touched that. I'm not sure if that's why appveyor failed.

src/cdi_CPEloss/Tabular_CP_Eloss.hh
KineticTheory: Another option (if you want to keep the namespace scope) would be `typedef std::experimental stdex;`. I think this is what the kokkos folks did in the test.

KineticTheory:
> @KineticTheory Could you take a look at the appveyor failure? There's a warning in there about unreachable code in a file in `c4` but Ben hasn't touched that. I'm not sure if that's why appveyor failed.

I've created a new branch 'cpeloss' that rebases this PR against `develop` and collapses into a single commit.  I will do a test build in my VS sandbox and then create a dummy-PR to see if appveyor still complains about unreachable code.

src/cdi/CPCommon.hh
```diff
@@ -20,9 +20,10 @@ namespace constants {
 unsigned int const num_CPModels(3);
 } // namespace constants
 
-enum class CPModel {
-  ELOSS = 0,        /*!< energy loss model */
-  LA_SCATTERING = 1 /*!< large-angle scattering model */
+enum class CPModelAngleCutoff {
+  NONE = 0, /*!< no angle cutoff */
+  TNBURN =
+      1 /*!< cutoff angle formula from "TN Burn Project: Minimum Impact parameter Scattering Angle Cutoff/Nuclear Interactions */
```
KineticTheory: Please wrap this comment to keep it below 80 cols.  See example at https://github.com/lanl/Draco/pull/723/files

src/cdi/CPEloss.hh
```diff
@@ -76,17 +78,18 @@ public:
...
    * \brief Query to determine the transporting particle type.
    */
-  virtual int32_t getProjectileZAID() const = 0;
+  //virtual int32_t getProjectileZAID() const = 0;
 
   /*!
    * \brief Query to determine the CP Model type.
    */
-  virtual rtt_cdi::CPModel getModel() const = 0;
+  //virtual rtt_cdi::CPModel getModel() const = 0;
```
KineticTheory: Do we need these pure virtual functions declarations?  Can these be removed?

src/ds++/DracoArray.hh
```diff
@@ -0,0 +1,135 @@
...
+    n3 = n3_in;
+    data.resize(n1 * n2 * n3);
+    allocated = true;
+  }
+  void zero() { std::fill(data.begin(), data.end(), 0); }
+};
+
+} // namespace rtt_dsxx
+
+#endif // rtt_dsxx_DracoArray_hh
```
KineticTheory: Remove DracoArray. Use mdspan instead.

src/ds++/DracoMath.hh
```diff
@@ -229,6 +229,62 @@ constexpr inline double linear_interpolate(double const x1, double const x2,
...
+ * \param[in] z    z coordinate of interpolation point
+ * \return The function value linearly interpolated to (x,y,z)
+ */
+inline double
+linear_interpolate_3(double const x0, double const x1, double const y0,
+                     double const y1, double const z0, double const z1,
+                     double const f000, double const f100, double const f001,
+                     double const f101, double const f010, double const f110,
+                     double const f011, double const f111, double const x,
+                     double const y, double const z) {
```
KineticTheory: As we discussed before, I really dislike this argument list as it will be error prone (17 doubles).  Consider reworking this algorithm to use `std::array<double,2>` for (x,y) pairs and maybe a 3x3 mdspan for f999.  I would like input from @along and @RyanWollaeger on ideas to make this interface less error prone.

RyanWollaeger: It seems the argument list could be made into three arrays: one size 6 for the lattice coordinates, one size 8 for the function, and one size 3 for the point.  Maybe the data array could be a `std::map` instead, then one could have, e.g., `data["f101"]` to access that function value.

brryan: I think arrays in the argument would be a good path forward, though I worry about performance even with `std::unordered_map`; this function may live inside heavy compute loops. From my discussion with @KineticTheory this morning, I think for now I will just move this code into the cdi_CPEloss library and leave the function signature as is -- it's not clear that there will be enough other uses for this code for it to be in ds++.

RyanWollaeger: Ah - that makes sense.  One could also define an enumeration, e.g., F000 = 0, F001 = 1, then access array values with the enumeration.

src/cdi_CPEloss/Tabular_CP_Eloss.hh
brryan: @KineticTheory `typedef std::experimental stdex;` didn't work for me, either at the namespace level or inside the class definition. What did work for me was `namespace stdex = std::experimental;` inside the `rtt_cdi_cpeloss` namespace in `Tabular_CP_Eloss.hh`; highlighting this here in case that is not OK.

KineticTheory:
Replaced by #725

PR #707, opened 2019-12-09
Optimize

kgbudge:
### Background

This is a quite minor tweak to Ordinate_Set that puts noexcept specifiers on some simple functions. I did this quite a while back and it's sat in my clone since. I figure it might as well go back upstream since it's correct, slightly useful, and minor.

### Description of changes

Add noexcept qualifier to several member functions of Ordinate_Set.

### Status

* Reference:
* Checks
  * [ ] Travis CI checks pass


src/quadrature/Ordinate_Set.hh
```diff
@@ -59,29 +59,31 @@ public:
   // ACCESSORS
 
   //! Return the ordinates.
-  vector<Ordinate> const &ordinates() const { return ordinates_; }
+  vector<Ordinate> const &ordinates() const noexcept { return ordinates_; }
```
KineticTheory: Marking this function `noexcept` seems like a really bad idea to me.   `ordinates_` may not even be allocated and yet you are returning the address to it.  I guess I need to be convinced that is is a good idea.  Am I missing a larger design or performance aspect related to this change?

From what I am reading about this pattern for _getters_ (this might even be an _antipattern_ but I'm not certain), they shouldn't be marked `noexcept` except in [cases where you can demonstrate an extreme performance benefit](https://stackoverflow.com/questions/21330807/should-i-use-noexcept-for-getters-always).

Also, I am [reading](https://stackoverflow.com/questions/10787766/when-should-i-really-use-noexcept) that 

> Most compilers follow a no-penalty-if-you-don't-throw exception handling method, so I doubt it would change much (or anything) on the machine code level of your code, although perhaps reduce the binary size by removing the handling code. 

So again, I'm not convinced that any of these changes will make a significant performance improvement and they add complexity where it isn't necessarily merited.

KineticTheory:
I also found this video useful and it leads me to think that the getters for POD types might be a good idea.

https://youtu.be/AG_63_edgUg

Again, I world like to have a better understanding of the use case and why you think noexcept is important here.

PR #731, opened 2020-01-30
Update build system to treat CUDA as a first class compiler.

KineticTheory:
### Background

+ Previously we treated CUDA as a third party vendor, but now that CMake can treat it as a real compiler, it made sense to update the build system to do the same.

### Purpose of Pull Request

* [Fixes Redmine Issue #1812](https://rtt.lanl.gov/redmine/issues/1812)

### Description of changes

+ Remove cuda setup instructions from `vendor_libraries.cmake`.  Compiler flags are now captured in the new file `unix-cuda.cmake` which is sourced from the `dbsSetupCuda` macro found in `compilerEnv.cmake`.  `dbsSetupCuda` should be called directly from a projects top level `CMakeLists.txt`.
+ Treat CUDA as an optional compiler in the same way that Fortran is treated as an optional compiler.  Update build system variables to mimic those used to setup Fortran:
  - `WITH_CUDA` --> `HAVE_CUDA`.
+ Properties for CUDA targets are now set in `component_macros.cmake` in the same way that C and C++ target properties are set.  We will default to requiring `CUDA_STANDARD=14` with vendor extensions not permitted. `CUDA_SEPARABLE_COMPILATION=ON` is defaulted.
+ Save CUDA related variables to `draco-config.cmake` and `ds++/config.h`:
  - `HAVE_CUDA`
  - `CUDA_DBS_STRING`
  - `Draco_CUDA_ARCH`
+ The new `unix-cuda.cmake` established compiler flags for nvcc.  This includes setting the `-arch` option.  The target arch value is saved so that client projects don't need to do a test compile to discover the hardware type.
  - Modify the configure-time test that determines the hardware features of the GPU so that no files are created in the source tree.
+ When calling `add_component_library` use the extra argument `LINK_LANGUAGE CUDA` to force `.cc` files to be compiled with the CUDA compiler. When the `LINK_LANUGUAGE=CUDA`, this cmake macros will add `-DUSE_CUDA`  to the compile options. Also, prefer creating `LIBRARY_TYPE STATIC` libraries.  Static libraries will not have the CUDA runtimes and all device symbols resolved and this prevents bizarre link errors related to `cmake_device_link.o`.
+ Update the configure summary so that CUDA information like the path to the compiler and the compiler flags are printed.
+ Removed redundant build variable `CUDA_DEVICE`.
+ Rework the `device` component to eliminate the `GPU_Device.cmake` files. 
+ Update `ds++/Assert.hh` to work correctly in a CUDA kernel (nothrow).

### Status

* Reference:
* Checks
  * [x] Travis CI checks pass


KineticTheory:
It looks like the Windows build is broken.  I'm looking into it...

config/CMakeAddFortranSubdirectory.cmake
```diff
@@ -250,16 +250,9 @@ function(cmake_add_fortran_subdirectory subdir)
     CONFIGURE_COMMAND ${CMAKE_COMMAND} -DCMAKE_BUILD_TYPE=${ep_build_type}
                       -P ${build_dir}/config_cafs_proj.cmake
     BUILD_COMMAND     ${CMAKE_COMMAND} -P ${build_dir}/build_cafs_proj.cmake
+    BUILD_ALWAYS 1
```
alexrlongne: Is this related to CUDA changes or something for debugging?

config/compilerEnv.cmake
```diff
@@ -596,6 +596,69 @@ macro(dbsSetupFortran)
...
+# - https://devblogs.nvidia.com/tag/cuda/
+# - https://devblogs.nvidia.com/building-cuda-applications-cmake/
+#------------------------------------------------------------------------------#
+macro(dbsSetupCuda)
+
+  # Toggle if we should try to build Cuda parts of the project.  This will be
+  # set to true if $ENV{FC} points to a working compiler.
+  option( HAVE_CUDA "Should we build Cuda parts of the project?" OFF )
+
+  # Is Fortran enabled (it is considered 'optional' for draco)?
```
alexrlongne: Change comment to CUDA

config/CMakeAddFortranSubdirectory.cmake
KineticTheory: This is related to an upstream fix in CMake that was a result of a bug I reported.  `CMakeAddFortranSubdirectory.cmake` is a highly modified version of a standard CMake capability. (Yes. I should submit some of my additions back to CMake.)

KineticTheory: Oh - in the MSVC builds, I need to use a pre-release of CMake to avoid a bug in CMake.  This change is required to allow the code to build correctly when using this pre-release.  It does not impact the Linux build.

config/compilerEnv.cmake
KineticTheory: Fixed in 140da76.

KineticTheory:
@alexrlongne I've addressed your comments and updated the PR. See 140da76.

alexrlongne:
Merging as only comments changed.

PR #744, opened 2020-02-24
CDI analytic DEDX upgrade

brryan:
### Background

* Refining stopping power support to be in line with target models

### Purpose of Pull Request

* Reorganize stopping powers so apart from the base stopping power class in CDI, everything lives in `CDI_CPEloss`
* Add stopping power of the form we probably want to use going forward, returning dE/dt for a projectile moving through a material composed of a single isotope with a given number density and temperature.

### Description of changes

* Moved analytic stopping power class to `CDI_CPEloss`
* Added Spitzer (coulomb logarithm is logarithm of debye length over classical distance of closest approach) analytic Eloss model

### Status

* Reference:
* Checks


src/cdi/CPCommon.hh
```diff
@@ -41,6 +41,14 @@ private:
 
 public:
   int32_t get_zaid() const { return zaid; }
+  int32_t get_z() const {
```
keadyk: It might be good to add an explanatory comment or two here, so future draco-ers know you're *intentionally* doing an integer division (and perhaps so they know about the special electron case!)

src/cdi_CPEloss/Analytic_Eloss_Model.hh
```diff
@@ -0,0 +1,84 @@
...
+ * This class is a pure virtual base class.
+ *
+ * The returned eloss coefficient is a power with units of keV shk^-1.
+ */
+//============================================================================//
+
+class Analytic_Eloss_Model {
+public:
+  // Typedefs.
+  typedef std::vector<char> sf_char;
```
keadyk: I'm sure I copy-pasted these typedefs in from some other class, but maybe it's dumb to have them here. I never know if typedef-ing in the base class is clever, or confusing...

src/cdi_CPEloss/Analytic_Spitzer_Eloss_Model.hh
```diff
@@ -0,0 +1,48 @@
+//----------------------------------*-C++-*-----------------------------------//
```
keadyk: Whoo hoo!! Now we have OPTIONS! :D

src/cdi_CPEloss/Analytic_Spitzer_Eloss_Model.hh
```diff
@@ -0,0 +1,48 @@
...
+/*!
+ * \file   cdi_CPEloss/Analytic_Spitzer_Eloss_Model.hh
+ * \author Ben R. Ryan
+ * \date   Feb 21 2020
+ * \brief  Analytic_KP_Alpha_Eloss_Model class definition.
```
keadyk: Looks like a copy-paste leftover :)

src/cdi_CPEloss/Analytic_Spitzer_Eloss_Model.cc
```diff
@@ -0,0 +1,69 @@
...
+
+  const double keV = 1e3 * pc.eV();
+  const double eloss_unit_fac = 1.e-8 / keV;
+
+  const double nt = rho / mt;
+  const double Tt = T * keV / pc.k();
+  const double vp = v * 1.e8;
+  const double vt = sqrt(2. * pc.k() * Tt / mt);
+  const double Ep = 1. / 2. * mp * vp * vp;
+  const double eloss_max = 1e3 * keV * nt;
```
keadyk: Is this cap specified by Spitzer, or Ben Ryan? Just curious :)

src/cdi_CPEloss/test/tstAnalytic_CP_Eloss.cc
```diff
@@ -107,9 +108,88 @@ void KP_alpha_test(rtt_dsxx::UnitTest &ut) {
...
-    FAILMSG("Analytic_CP_Eloss test fails.");
+    FAILMSG("KP_Alpha CPEloss test fails.");
+
+  return;
+}
+
+void Spitzer_test(rtt_dsxx::UnitTest &ut) {
+
+  // Deuterium:
+  int deuterium_zaid = 1002;
```
keadyk: Even though this is just a test, it might be good to explicitly use int32_t here to match the signature!

src/cdi_CPEloss/test/tstAnalytic_CP_Eloss.cc
```diff
@@ -107,9 +108,88 @@ void KP_alpha_test(rtt_dsxx::UnitTest &ut) {
...
+  double deuterium_mass = 3.34358e-24;
+  rtt_cdi::CParticle target_in(deuterium_zaid, deuterium_mass);
+  // Alpha particle:
+  int alpha_zaid = 2004;
```
keadyk: Same as above :)

src/cdi/CPCommon.hh
brryan: Good point -- done

src/cdi_CPEloss/Analytic_Eloss_Model.hh
brryan: Good catch -- these aren't used currently so I just took them out

src/cdi_CPEloss/Analytic_Spitzer_Eloss_Model.hh
brryan: One thing to maybe worry about is that by adding options I changed the analytic interface such that it now takes in `target` and `projectile` `CParticle` objects. I *think* this is a good way to handle stopping powers going forward, e.g. we get dE/dt separately for each isotope in a target material, but I don't know that the KP alpha model (fitting formula for alphas moving through a combined DT gas) can be easily reconciled to this. So maybe KP alpha should come out at some point? Maybe it is still useful to have, I don't know.

src/cdi_CPEloss/Analytic_Spitzer_Eloss_Model.hh
brryan: That'll teach me to scrub wikipedia for stopping power source code...

src/cdi_CPEloss/Analytic_Spitzer_Eloss_Model.cc
brryan: I just copied this from what's used for a practically-infinite stopping power in some tabulated data -- since the units don't make sense I'll define a new `rtt_cdi::constants::max_eloss` up in the `CPCommon` to be used throughout.

src/cdi/CPCommon.hh
```diff
@@ -41,6 +44,16 @@ private:
...
+    if (zaid != -1) {
+      // Integer division to extract the ZZZ from ZAID format ZZZAAA
+      return zaid / 1000;
+    } else {
+      // Electrons have special ZAID = -1
+      return -1;
```
KineticTheory: Can you add a unit test that captures this branch?

brryan: Done (test for new TR stopping power model tests both ion and electron cases, which differ)

src/cdi/CPCommon.hh
```diff
@@ -18,11 +18,14 @@ namespace rtt_cdi {
 namespace constants {
 //! Number of models contained in rtt_cdi::CPModel.
 unsigned int const num_CPModels(3);
+
+//! "Effectively infinite" stopping power (keV sh^-1)
+constexpr double max_eloss = 1.e15;
```
keadyk: Nice, I like this! Thanks for adding it.

src/cdi_CPEloss/Analytic_Spitzer_Eloss_Model.hh
keadyk: That's a good point. I think KP Alpha can definitely come out down the road -- it was just the dumbest, most bare-bones first option I found in Edward's write-up. Once we've used it for preliminary comparisons, I'm happy to ditch it :)

PR #754, opened 2020-03-02
fix xthi ythi for apple builds

clevelam:
### Background

* Similar to windows systems the hardware architecture needs to be queried differently the linux. This PR supplies the necessary information to do so.  

### Purpose of Pull Request

* Fix xthi and ythi issues on apple
* [Fixes xthi ythi issue on apple #1849](https://rtt.lanl.gov/redmine/issues/1849)

### Status

* Reference:
* Checks

### DIFF
--- a/src/c4/xthi_cpuset.hh
+++ b/src/c4/xthi_cpuset.hh
@@ -54,7 +54,7 @@ namespace rtt_c4 {
  *
  * \return A string of the form "0-8;16-32;" or "0-63"
  */
-#if APPLE
+#ifdef APPLE
 
 #include <sys/sysctl.h>
 #include <sys/types.h>
@@ -79,19 +79,20 @@ namespace rtt_c4 {
 #define SYSCTL_CORE_COUNT "machdep.cpu.core_count"
 
 typedef struct cpu_set {
-  uint32_t count;
+  uint64_t count;
 } cpu_set_t;
 
 static inline void CPU_ZERO(cpu_set_t *cs) { cs->count = 0; }
 
 static inline void CPU_SET(int num, cpu_set_t *cs) { cs->count |= (1 << num); }
 
-static inline int CPU_ISSET(int num, cpu_set_t *cs) {
-  return (cs->count & (1 << num));
+static inline bool CPU_ISSET(int num, cpu_set_t *cs) {
+  uint64_t teh_bit = 1ull << num;
+  return (cs->count & teh_bit) != 0;
 }
 
 int sched_getaffinity(pid_t pid, size_t cpu_size, cpu_set_t *cpu_set) {
-  int32_t core_count = 0;
+  int64_t core_count = 0;
   size_t len = sizeof(core_count);
   int ret = sysctlbyname(SYSCTL_CORE_COUNT, &core_count, &len, 0, 0);
   if (ret) {
--- a/src/c4/xthi_cpuset.hh
+++ b/src/c4/xthi_cpuset.hh
@@ -25,6 +25,12 @@
 #include <processthreadsapi.h> // requries SystemCall.hh to be loaded first.
 #endif
 
+/*! \bug Broken on OSX (sched_getaffinity, cpu_set_t, etc. not defined).
+ *
+ * Possible solution at
+ * https://yyshen.github.io/2015/01/18/binding_threads_to_cores_osx.html
+ */
+
 namespace rtt_c4 {
 
 //----------------------------------------------------------------------------//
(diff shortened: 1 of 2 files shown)


src/c4/xthi_cpuset.hh
```diff
@@ -48,6 +54,60 @@ namespace rtt_c4 {
  *
  * \return A string of the form "0-8;16-32;" or "0-63"
  */
+#if APPLE
```
KineticTheory: Change this to `#ifdef APPLE` to fix the build failure reported by Travis.

src/c4/xthi_cpuset.hh
```diff
@@ -48,6 +54,60 @@ namespace rtt_c4 {
...
+
+/*
+ * I think this is the mac equivelent of CPU_SETSIZE
+ *
+ * example - https://stackoverflow.com/questions/150355/programmatically-find-the-number-of-cores-on-a-machine
+ * documentation - https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/sysconf.3.html
+ *
+*/
+#ifndef CPU_SETSIZE
+#define CPU_SETSIZE _SC_NPROCESSORS_ONLN
```
timmah: When I look at CPU_SETSIZE on my laptop, I get 58. Does this mean that the integer in cpu_set needs to accomodate at least 58 bits (i.e. be a 64b integer)? 

I ask because I'm seeing a funny result: if I call `rtt_c4::sched_getaffinity` directly, I get coremask equal to 15, or 0xF, which is just the four lowest bits set (1+2+4+8; this is on a quad-core mac, eh)? But when I call `rtt_c4::cpuset_to_string`, I get core affinity 0-3,32-35. If I change to uint64_t in cpu_set (line 82) and int64_t in sched_getaffinity (line 94), then I get 0-3. 

So I'm wondering if what's happening is the bit shift is wrapping around, and the call to CPU_ISSET at line 165 triggers twice, because it's iterating the same 32 bits twice.

timmah:
@KineticTheory  There's a related bug that I just found: xthi assumes OpenMP is present, and is compiled even if OpenMP is not found by CMake. Some possible fixes:
1. Make OpenMP required in CMake
1. Don't build xthi if OpenMP is not found by CMake
1. Wrap OMP calls in xthi with `#ifdefs` and sensible defaults. 

Do you want to fix that on this PR, or start another?

clevelam:
I'll try and get some help from @KineticTheory to incorporate the OpenMP issues into this PR.

src/c4/xthi_cpuset.hh
clevelam: So I get core affinity = 0-5,32-37 regardless if I make these changes or leave them the same. I have a Sandy Bridge-E on my mac (I think this is 6 cores). Thoughts?

timmah: I also changed CPU_ISSET. Here are the relevant diffs that I have:
```
@@ -79,19 +79,21 @@ namespace rtt_c4 {
 #define SYSCTL_CORE_COUNT "machdep.cpu.core_count"

 typedef struct cpu_set {
-  uint32_t count;
+  uint64_t count;
 } cpu_set_t;

 static inline void CPU_ZERO(cpu_set_t *cs) { cs->count = 0; }

 static inline void CPU_SET(int num, cpu_set_t *cs) { cs->count |= (1 << num); }

-static inline int CPU_ISSET(int num, cpu_set_t *cs) {
-  return (cs->count & (1 << num));
+static inline bool CPU_ISSET(int64_t num, cpu_set_t *cs) {
+  uint64_t teh_bit = 1ull << num;
+  return (cs->count & teh_bit) != 0;
 }

-int sched_getaffinity(pid_t pid, size_t cpu_size, cpu_set_t *cpu_set) {
-  int32_t core_count = 0;
+int sched_getaffinity(pid_t /*pid*/, size_t /*cpu_size*/, cpu_set_t *cpu_set) {
+  int64_t core_count = 0;
```

clevelam: That seems to have fixed it. Now I get core affinity = 0-5.

clevelam: ythi returns:
Found 12 logical CPUs (hardware therads) per node.
deschutes.lanl.gov :: Rank 00000, Thread 000, core affinity = 0-5

timmah: Yeah, there's ambiguity about whether we're talking about physical cpus or hyperthreads. So in xthi, we're calling your `sched_getaffinity`, which queries physical processor cores. In ythi, we're calling `std::thread::hardware_concurrency`, which seems to go with hyperthreads, at least on Mac. (cppreference.com includes this gem of clarity in the description of hardware_concurrency: "The value should be considered only a hint." ) But `cpuset_to_string` is still using `sched_getaffinity` and physical cores, so it only reports 6 affinity slots. We might consider changing ythi to report consitently with the affinity mask and ignore hyperthreads, or be more clear about what's being reported?

KineticTheory:
I'll let you and @timmah tell me if the code works on APPLE or not.  Do we need to rework the logic in xthi and ythi so they are both talking about physical cores vs hyperthreads?

PR #793, opened 2020-04-03
Apply recommended clang-tidy transformations to c4.

KineticTheory:
### Background

+ Only examine `-checks=modernize-*` checks but ignore `-modernize-use-trailing-return-type` since this is just ugly IMHO.

### Description of changes

+ Apply recommended updates...
+ Change `typedef` commands to `using myalias=actual`.
+ Use `auto` when assigning to avariable with a `static_cast`.
+ Replace C-style arrays with `std::array`.
+ Replace `NULL` with `nullptr`.
+ Provide default initial values for class members and local variable declarations.
+ Use `override` in place of `virtual` for member functions in a class inheritance.
+ Disable move/copy asignment and move/copy constructors for some classes.  If these were in the `private` section of a class definition, move them to the public section.
+ Mark constructor and destructors `=default` when body is empty.
+ Fix a bug in `tstAbort.cc` related to unmatched brackets. I'm not sure how this compiled previously.
+ For tests that pull in the entire `std::` namespace, remove `std::` from functions and containers.
+ Make better use of `UnitTest` macros like `FAIL_IF_NOT`.
+ Use modern `for-each` style for-loops.
+ Rewrap and indent comment blocks to use 80 character width.  Remove some outdated comments. Made some spelling fixes.

### WIP

+ [x] Visual Studio Build failing
+ [x] gcc-8.3 build is failing
+ Commentary on questions below (turn off check, or fix it)
   + [x] No matching nonblocking call.
   + [x] Use of memory after it is freed
   + [x] Value stored to 'i' is never read

### Need help.

+ There are some warnings that I need some help with:
  + @alexrlongne @timmah @kgbudge Can you help with any of these.

```console
/home/kellyt/draco/src/c4/C4_Req.cc:125:5: warning: Request  has no matching nonblocking call.  [clang-analyzer-optin.mpi.MPI-Checker]
    MPI_Wait(&r, s);
    ^
/home/kellyt/draco/src/c4/C4_Req.cc:119:7: note: Assuming the condition is true
  if (assigned) {
      ^
/home/kellyt/draco/src/c4/C4_Req.cc:119:3: note: Taking true branch
  if (assigned) {
  ^
/home/kellyt/draco/src/c4/C4_Req.cc:121:9: note: Assuming 'status' is null
    if (status) {
        ^
/home/kellyt/draco/src/c4/C4_Req.cc:121:5: note: Taking false branch
    if (status) {
    ^
/home/kellyt/draco/src/c4/C4_Req.cc:125:5: note: Request  has no matching nonblocking call. 
    MPI_Wait(&r, s);
    ^
```

```console
/home/kellyt/draco/src/c4/C4_Req.hh:126:31: warning: Use of memory after it is freed [clang-analyzer-cplusplus.NewDelete]
  bool inuse() const { return p->inuse(); }
                              ^
/home/kellyt/draco/src/c4/C4_Req.cc:57:17: note: Assuming req == *this
C4_Req &C4_Req::operator=(const C4_Req &req) {
                ^
/home/kellyt/draco/src/c4/C4_Req.cc:58:3: note: Calling 'C4_Req::free_'
  this->free_();
  ^
/home/kellyt/draco/src/c4/C4_Req.cc:78:7: note: Assuming the condition is true
  if (p->n <= 0)
      ^
/home/kellyt/draco/src/c4/C4_Req.cc:78:3: note: Taking true branch
  if (p->n <= 0)
  ^
/home/kellyt/draco/src/c4/C4_Req.cc:79:5: note: Memory is released
    delete p;
    ^
/home/kellyt/draco/src/c4/C4_Req.cc:58:3: note: Returning; memory was released
  this->free_();
  ^
/home/kellyt/draco/src/c4/C4_Req.cc:60:7: note: Calling 'C4_Req::inuse'
  if (req.inuse())
      ^
/home/kellyt/draco/src/c4/C4_Req.hh:126:31: note: Use of memory after it is freed
  bool inuse() const { return p->inuse(); }
                              ^
```

```console
/home/kellyt/draco/src/c4/test/tstOMP.cc:215:5: warning: Value stored to 'i' is never read [clang-analyzer-deadcode.DeadStores]
    for (int i = 0; i < N; ++i) {
    ^
/home/kellyt/draco/src/c4/test/tstOMP.cc:215:5: note: Value stored to 'i' is never read
/home/kellyt/draco/src/c4/test/tstOMP.cc:215:10: warning: Value stored to '.omp.iv' is never read [clang-analyzer-deadcode.DeadStores]
    for (int i = 0; i < N; ++i) {
```

### Status

* Reference:
* Checks


src/CMakeLists.txt
```diff
@@ -51,12 +51,13 @@ message( STATUS "Configuring Level 1 packages --" )
...
 
 # Only test ds++ if CI_CLANG_TIDY is running
-if( NOT CI_CLANG_TIDY )
 
 # Level 2
 message(" ")
 message( STATUS "Configuring Level 2 packages --" )
 add_dir_if_exists( c4 )           # needs ds++
+
+if( NOT CI_CLANG_TIDY )
```
KineticTheory, line 60: Instruct the nightly clang-tidy regression that runs on ccscs1 to omit checking all directories listed after this line.

src/c4/C4_Req.cc
```diff
@@ -97,28 +97,13 @@ void C4_ReqRefRep::free() {
...
- * \brief Destructor.
- *
- * It is important that all existing requests are cleared before the destructor
- * is called.  We used to have a wait() in here; however, this causes exception
- * safety problems.  In any case, it is probably a bad idea to clean up
- * communication by going out of scope.
- */
-//----------------------------------------------------------------------------//
-C4_ReqRefRep::~C4_ReqRefRep() { /* empty */
-}
```
KineticTheory, line 120: Moved the the `.hh` file and marked `=default`.

KineticTheory:
Evidently gcc-7 has different behavior than gcc-8.  I had to install and use gcc-8 to reproduce and fix the build errors reported by the CI.

timmah:
That first warning is interesting. Am I reading correctly that it is only pointing out the nonexistence of a nonblocking `MPI_Wait`? 

I think the second warning is maybe valid. That is some interesting code. I wonder if it's from pre-smart pointer days? Looks like `C4_Req` is wrapping a heap allocated, reference-counted pointer to the actual implementation ("letter/envelope idiom"). Maybe the call to `p->inuse()` should be guarded by `Insist(p != nullptr)`?

Is the third one a false positive? https://clang-analyzer.llvm.org/faq.html

KineticTheory:
@timmah Thanks for the suggestions.  I agree that (3) is a false positive.  I think the tips in the web page you sent will help me fix that issue.  I'll also try the `Insist` for (1) to see if any of our tests trip on it.

KineticTheory:
False positive (3) is documented here: https://bugs.llvm.org/show_bug.cgi?id=37442

timmah:
tl;dr What Alex said!

Not sure the MPI checker is going to do a lot for you. It looks like it can check three things:

1. Have you used one request in two async calls without waiting on the request in before the second call?
2. Have you made an async call but not made the corresponding wait?
3. Are you waiting on a request that was never used in an async call?

The problem is that libtooling typically only works on one translation unit at a time. Our MPI calls are in separate TUs (`C4_Req.cc` and `C4_MPI_async_pt.cc`), and the code that calls them is in separate libraries getting linked into various applications. 

So even if you ran the MPI checker on say one of the `IMC_Manager_*_pt.cc`, the MPI calls themselves are in the TUs for C4. The checker is never going to see the MPI calls coming out of that TU: that control flow is only available at link time but the checker works early at compile time (preprocessor and AST). So you might get a false sense of security that the checks are not firing.

And since the `MPI_Wait` is in C4_Req.cc and the async calls are in C4_MPI_async_pt.cc, they're never going to match up either--you'll only ever get false positives. I may take a look tomorrow and see if there's some magic cross-TU thing happening here, but I doubt it.

KineticTheory:
@timmah What are you recommendations for suppressing these clang-analyzer warnings?  It isn't obvious to me that these checks can be disabled.  Even the `#ifdef` didn't seem to work as advertised.

KineticTheory:
For case (2), I added an `assert(p!=nullptr);` and `#ifndef __clang_analyzer__`.  This prevents clang-analyze from giving a false positive while ensuring that `p->inuse()` is a valid call.

KineticTheory:
I have trinitite build failures.  I need to apply the same `#ifdef` I use for MSMPI in `tstComm_Dup.cc`.  ☹️ 

I thought MPI had an API that everyone had to adhere to.  I guess I was mistaken.  I need `MPI_Comm new_comm = nullptr;` for OpenMPI, but `MPI_Comm new_comm = 0;` for MSMPI and MPICH.

keadyk:
I have a GitHub mechanics question: Tim references comments by Alex, but I see no such comments. Do conversations tied to lines of code disappear when you force-push to a PR branch? I do this occasionally myself, so I am curious to know the answer :)

alexrlongne:
@keadyk Sorry, that's on me. I replied to an e-mail instead of posting my comments here. I'll post my e-mail below!

>KT and Tim,
 
>I looked at the clang-tidy documentation and code for this call and I can't tell why clang-tidy >thinks it can reason about non-blocking calls associated with this wait operation. The purpose of >this warning is to make sure you're not calling "wait" with a request that's never been associated >with some non-blocking call ("Isend" for instance). We use the "assigned" flag to ensure we only >wait if a request is active so I'm not worried about this warning. If it was coming from a function >where this was used I would definitely investigate. All that said, I don't know a lot about how >clang AST works so maybe Tim has a better idea about this error.
 
>Alex

keadyk:
@alexelongne Ohhhhh! Okay, fair enough :)

KineticTheory:
b26ceec fixed the trinitite build issue.  All tests are passing. This is ready for final review.

src/c4/C4_MPI.hh
```diff
@@ -85,7 +85,7 @@ void broadcast(ForwardIterator first, ForwardIterator last,
   Remember(int check =) broadcast(&size, 1, 0);
   Check(check == MPI_SUCCESS);
 
-  value_type *buf = new value_type[size];
+  auto *buf = new value_type[size];
   if (node() == 0)
```
timmah, line 89: Hmmm, is replacing a type with `auto` one of the new best practices? I think this tends to obscure code.

KineticTheory: `auto` is only recommended in specific cases where the type is clearly identified on the RHS.  (1) when using `new`, and (2) when assigning via `static_cast`. Examples:

```c++
// Case 1
auto *buf = new int[size]`

// Case 2
auto bar = static_cast<double>(get_data());
```

timmah: Got it.

keadyk:
On it!

src/c4/C4_MPI.hh
keadyk: Interesting! That's a logical use case :)

src/c4/Timer.cc
```diff
@@ -347,8 +346,8 @@ void Timer::printline_mean(std::ostream &out, unsigned const p,
   double s = sum_system_cpu(), s2 = s * s;
```
keadyk, line 346: @KineticTheory I know you didn't touch this line or the next, but isn't this (multiple initializations on a line) frowned upon in our style guide?

src/c4/opstream.hh
```diff
@@ -76,7 +76,7 @@ private:
     void send();
     void shrink_to_fit();
 
-    virtual int_type overflow(int_type c);
+    int_type overflow(int_type c) override;
```
keadyk, line 79: Nice! I had almost forgotten override existed!

src/c4/Termination_Detector.cc
```diff
@@ -54,16 +54,18 @@ void Termination_Detector::init() {
...
 }
+#else
+Termination_Detector::~Termination_Detector() = default;
+#endif
 
 //----------------------------------------------------------------------------//
 //! Return \c true if we have terminated; \c false otherwise.
 bool Termination_Detector::is_terminated() {
-  static unsigned buffer[3];
+  static std::array<unsigned, 3> buffer;
```
keadyk, line 68: Is 3 considered "magic number" code smell in this context, or is it okay?

keadyk:
This LGTM! My comments are minor, and mostly for my own clarification.

src/c4/Termination_Detector.cc
KineticTheory: Created #795 to address this.

src/c4/Timer.cc
KineticTheory: See #795

KineticTheory:
@keadyk You can merge this when you are ready.  I created issue #795 to address a couple of your comments.  That will be done in a separate PR.

keadyk:
@KineticTheory Sounds good, merging now!

PR #808, opened 2020-04-21
Fix quo test failure by treating std::thread as upper resource bound

alexrlongne:
### Background

* A Quo test was failing on RZ Ansel that checked to see if number of available hardware threads that Quo sees matched the value of `std::thread::hardware_concurrency` (https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency). This difference is because Quo uses an hwloc command that gets the number of cores that are available for use (as set by the system admins) where as std::thread seems to just see the total number of cores on the machine (I don't know exactly how it gets this but its number matches the value in /proc/cpuinfo). We would like to only use the available cores and thus go with the answer Quo gives. I've modified the test to treat the `std::thread::hardware_concurrency` value as an upper bound.

### Purpose of Pull Request

* [Fixes Redmine Issue #1879](https://rtt.lanl.gov/redmine/issues/1879)

### Description of changes

* Uncomment test line checking hardware thread counts.
* Add description of quo and std::thread behavior for detecting hardware thread numbers

### Status

* Reference:
* Checks


KineticTheory:
LGTM.

PR #818, opened 2020-05-01
Change CMake logic for lib cdi_ndi to ensure upstream code can build when NDI is not found

keadyk:
### Background

* previously, the cdi_ndi library wasn't built *at all* when NDI was not found
* this is fine in theory, but it made things tricky in upstream code; because the NDI_FOUND cmake variable was not available, there was no obvious way to protect against includes of draco cdi_ndi header files that were not installed.

### Purpose of Pull Request

* Instead of omitting the cdi_ndi lib entirely when NDI is not found, stub out the routines that need NDI (have them throw assertions)
* This means that upstream code can be built without protecting cdi_ndi includes in macros -- the calls to cdi_ndi will just throw if it is not found
* This is needed to get a corresponding pull request in gitlab going (I will update the description of that PR shortly)

### Description of changes

* update CMakeLists logic for cdi_ndi - there was previously a confusing "if(NDI_FOUND) and not environment(travis)" statement wrapping the cmake logic. I think it was trying to protect against building the cdi_ndi lib in CI -- but you'd think that would be automatic because NDI_FOUND would be false... (?!)
* Move ndi.h include into config file, and remove from other cdi_ndi files
* Wrap actual calls to ndi routines / references to env. variables in NDI_FOUND macro
* @KineticTheory I am open to other ways to do this in CMake, and also more than willing to explain why I did things this way :)

### Status

* Reference:
* Checks

### DIFF
--- a/src/cdi_ndi/CMakeLists.txt
+++ b/src/cdi_ndi/CMakeLists.txt
@@ -51,9 +51,13 @@ add_component_library(
    LIBRARY_NAME cdi_ndi
    SOURCES      "${sources}"
    HEADERS      "${headers}" )
-target_include_directories( Lib_cdi_ndi PUBLIC ${NDI_INCLUDE_DIR}
+
+target_include_directories( Lib_cdi_ndi
    PUBLIC $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}> )
 
+if(TARGET NDI::ndi)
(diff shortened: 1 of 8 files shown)


KineticTheory:
@keadyk I'll review this over the weekend.

keadyk:
@KineticTheory Thank you -- no rush!! Hopefully my latest commit fixes the last few warnings :)

PR #832, opened 2020-05-13
Reorder enumeration of geometry in mesh_element/Geometry.hh

kgbudge:
The new enumeration is consistent with common practice in hydrodynamics codes,
where the enumerated value is the number of suppressed dimensions. See the
expanded description of the enumeration in the modified definition for further
explanation.

By adopting an enumeration that matches many hydrocodes, we avoid having to do as much translation from host code enumeration to our library enumeration, with all the unpleasant surprises that brings.

### Description of changes

* Reorder enumerators without changing their names. The corrensponding integral value for each enumerator then matches the common hydrodynamics code definition.

### Status

* Reference:
* Checks

### DIFF
--- a/src/mesh_element/Geometry.hh
+++ b/src/mesh_element/Geometry.hh
@@ -13,11 +13,19 @@
 
 namespace rtt_mesh_element {
 
-//! Enumerates supported geometries.
+/*! Enumerates supported geometries.
+ *
+ * The order of enumerated values is not arbiterary. The corresponding integral
+ * value is the number of suppressed dimensions in the geometry, e.g., axisymmetric
+ * geometry looks 2-D but is actually 3-D (one suppressed dimension) while spherical
+ * geometry looks 1-D but is actually 3-D (two suppressed dimensions.) The number
+ * of suppressed dimensions is used in some formulas in a number of hydrodynamics
+ * codes, so it seems like a good idea for us to adopt this convention as well.
+ */
 enum Geometry {
+  CARTESIAN,    //!< 1D (slab) or 2D (XY) Cartesian geometry
   AXISYMMETRIC, //!< 2D (cylindrical) R-Z
   SPHERICAL,    //!< 1D SPHERICAL
-  CARTESIAN,    //!< 1D (slab) or 2D (XY) Cartesian geometry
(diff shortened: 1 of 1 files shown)


src/mesh_element/Geometry.hh
```diff
@@ -13,11 +13,19 @@
...
+ * geometry looks 1-D but is actually 3-D (two suppressed dimensions.) The number
+ * of suppressed dimensions is used in some formulas in a number of hydrodynamics
+ * codes, so it seems like a good idea for us to adopt this convention as well.
+ */
 enum Geometry {
+  CARTESIAN,    //!< 1D (slab) or 2D (XY) Cartesian geometry
   AXISYMMETRIC, //!< 2D (cylindrical) R-Z
   SPHERICAL,    //!< 1D SPHERICAL
-  CARTESIAN,    //!< 1D (slab) or 2D (XY) Cartesian geometry
   END_GEOMETRY  //!< Sentinel value
```
attom, line 29: Do we want to be explicit: 

```
{
CARTESIAN = 0,
AXISYMMETRIC = 1,
SPHERICAL = 2,
END_GEOMETRY
}
```

KineticTheory: > Do we want to be explicit:
> 
> ```
> {
> CARTESIAN = 0,
> AXISYMMETRIC = 1,
> SPHERICAL = 2,
> END_GEOMETRY
> }
> ```

Yes, and it would also be better if this was an `enum class`.  I've already merged this PR, but feel free to create a new one.

kgbudge: Making it explicit seems okay.
Doesn't making it an enum class prohibit conversion to/from int? Which kind of obviates the whole point.

KineticTheory: The whole point is exactly that   An enum should not have an easily conversion path to integers.  This enforces strong type checking. See http://www.stroustrup.com/C++11FAQ.html#enum.

kgbudge: With any other enumeration, that would make sense.

The whole point of this merge is that the integer value actually means something significant, in and of itself.

PR #857, opened 2020-06-19
Add parallel ensight capability

clevelam:
### Background

* The previous implementation of the ensight viz package only accounted for domain replicated problems. This commit enables domain decomposed ensight files using the ofpstream parallel file writer.

### Purpose of Pull Request

* Add domain decomposed ensight write capability

### Description of changes

* Add domain_decomposed option to Ensight_Stream and Ensight_Translator
* Add a decomposed_stream and a serial_stream file options to Ensight_Stream
* Set Ensigh_Stream pointer based on the decomposition flag
* Add a flush function to Ensight_Stream
* Update Ensight_Translator to collect and set offsets for domain_decomposed problems

### Status

* Reference:
* Checks


clevelam:
@KineticTheory I think this is ready. I'm tripping on the new lint python check and the code coverage went up but didn't meet the 94.2% because of 2 dbc that aren't covered in the testing.

src/viz/Ensight_Stream.cc
```diff
@@ -71,19 +73,33 @@ Ensight_Stream::~Ensight_Stream(void) { close(); }
...
+    if (binary)
+      d_decomposed_stream.reset(
+          new rtt_c4::ofpstream(file_name, std::ios::binary));
+    else
+      d_decomposed_stream.reset(new rtt_c4::ofpstream(file_name));
+    // set to a generic ostream
+    d_stream = &*d_decomposed_stream;
+  } else {
+    Insist(rtt_c4::node() == 0, "Ensight_Stream, called by nonzero rank with "
+                                "the domain decomposed flag");
```
keadyk, line 95: Is this error message accurate? I thought this threw because a nonzero rank was calling Ensight Stream withOUT the domain-decomposed flag :)

src/viz/Ensight_Translator.i.hh
```diff
@@ -434,58 +457,92 @@ void Ensight_Translator::write_geom(const uint32_t part_num,
...
+
+  // set parallel offset for cell and vertices numbers
+  if (d_decomposed)
+    rtt_c4::global_sum<int>(&node_verts[0], rtt_c4::nodes());
+  int local_vert_offset = 0;
+  for (auto n = 0; n < rtt_c4::nodes(); n++) {
+    if (n < rtt_c4::node()) {
+      local_vert_offset += node_verts[n];
+    }
+  }
```
keadyk, line 492: Could this logic to get the vert_offsets maybe be replaced with a prefix_sum() from c4? I used these to calculate rank-wise global cell offsets in the parmetis PR if you want to look at an example. They're declared in src/c4/C4_Functions.hh :)

KineticTheory:
@clevelam I know this is a lot to ask - but would you mind merging #860 and rebasing this PR.  I think #860 will fix the linter failure

src/viz/Ensight_Translator.i.hh
```diff
@@ -49,19 +50,23 @@ template <typename SSF>
...
     : d_static_geom(static_geom), d_binary(binary), d_dump_dir(gd_wpath),
       d_num_cell_types(0), d_cell_names(), d_vrtx_cnt(0), d_cell_type_index(),
       d_dump_times(), d_prefix(), d_vdata_names(vdata_names),
       d_cdata_names(cdata_names), d_case_filename(), d_geo_dir(),
       d_vdata_dirs(), d_cdata_dirs(), d_geom_out(), d_cell_out(),
-      d_vertex_out() {
+      d_vertex_out(), d_decomposed(decomposed) {
+  // Insist(!d_static_geom,
+  //        "We no longer support static geometry (only one geometry file "
+  //        "created per time step) just set this to false");
```
keadyk, line 63: What is this guy about? :)

src/viz/Ensight_Stream.cc
```diff
@@ -43,12 +44,13 @@ Ensight_Stream &endl(Ensight_Stream &s) {
  * \param file_name  Name of output file.
  * \param binary     If true, output binary.  Otherwise, output ascii.
  * \param geom_file  If true, then a geometry file will be dumped.
+ * \param decomposed If true, input is domain decomposed. Otherwise domain replicated.
  */
 Ensight_Stream::Ensight_Stream(const std::string &file_name, const bool binary,
-                               const bool geom_file)
-    : d_stream(), d_binary(false) {
+                               const bool geom_file, const bool decomposed)
+    : d_decomposed_stream(), d_serial_stream(), d_stream(), d_binary(false) {
```
keadyk, line 51: Am I the only one who thinks it's weird that we set d_binary to false in the initializer list, even though the binary parameter is passed to the constructor? It looks like we set d_binary in the open() call instead. (I know you didn't write that code, Matt... just something I noticed when you changed code in the vicinity :) )

keadyk:
Thanks for doing this, @clevelam! LGTM :)

src/viz/Ensight_Stream.cc
clevelam: Good catch. I'll fix this.

src/viz/Ensight_Translator.i.hh
clevelam: Cool. I will try this. I didn't know such a function existed.

src/viz/Ensight_Translator.i.hh
clevelam: Ah... I was going to prevent this option, but decided to leave it in. I'll remove it.

src/viz/Ensight_Stream.cc
clevelam: Yes, this is strange. I had intended to fix this and got lost in pointer world so I forgot... I'll fix it.

clevelam:
> @clevelam I know this is a lot to ask - but would you mind merging #860 and rebasing this PR. I think #860 will fix the linter failure

No problem. I'll just incorporate your changes if that is alright. Then if all looks good we can close your PR.

src/viz/Ensight_Stream.cc
KineticTheory: I would seem that we should use a default argument value if the default value should be defaulted.
```c++
Ensight_Stream::Ensight_Stream(const std::string &file_name, const bool binary=faluse,
 const bool geom_file, const bool decomposed)
    : d_decomposed_stream(), d_serial_stream(), d_stream(), d_binary(binary) {
```

src/viz/Ensight_Stream.hh
```diff
@@ -52,8 +55,14 @@ private:
 
   // DATA
 
-  // The actual file stream.
-  std::ofstream d_stream;
+  // An optional parallel stream for domain decomposed geometry
```
KineticTheory, line 58: Can you change the comments to `//!` so doxygen can see them?

src/viz/test/CMakeLists.txt
```diff
@@ -21,12 +21,13 @@ file( GLOB test_sources
 # Build Unit tests
 # ---------------------------------------------------------------------------- #
 
-if( NOT APPLE )
+if( NOT APPLE AND DRACO_C4 STREQUAL MPI )
```
KineticTheory, line 24: Nice!  Since the CI tests the `DRACO_C4=SCALAR`, I'll assume this change is safe.

src/viz/test/CMakeLists.txt
```diff
@@ -52,12 +53,14 @@ if( NOT APPLE )
     DRIVER ${CMAKE_CURRENT_SOURCE_DIR}/tstEnsight_Translator.py
     APP    $<TARGET_FILE_DIR:Ut_viz_tstEnsight_Translator_exe>/$<TARGET_FILE_NAME:Ut_viz_tstEnsight_Translator_exe>
     BUILDENV "CMAKE_HOST_SYSTEM_PROCESSOR=${CMAKE_HOST_SYSTEM_PROCESSOR}"
+    PE_LIST "1;2;4"
```
KineticTheory, line 56: Do we need to add a `RESOURCE_LOCK` to ensure that these tests don't access in/out files simultaneously?

src/viz/Ensight_Stream.cc
clevelam: So this is tricky. The set doesn't occur until open (which is a public call), which is only optionally called. Maybe I should ditch this constructor, and force the open function to be called independently. The way this is currently written is clunky. What do you all think?

clevelam: I think ofstream is a little like this (optional open in construction), though the full details I do not know for sure.

keadyk: Hmmm... Even though the setting of d_binary currently doesn't happen til the open() function, would it actually break other code to have it set in the constructor's initializer list? It seems like it wouldn't hurt to have it set for the Ensight_Stream object, even if the open() function isn't called on construction. The binary flag could be taken out of the open() arguments list, with the understanding that the open() function could only be called on the type of file the Ensight_Stream object was created for (binary or ascii).

Sorry if there is something I'm missing here! I'm also fine to leave it alone, since you didn't introduce it and thus shouldn't have to waste half your life trying to fix it :)

clevelam: I opted to set d_binary with binary, no additional default, because this was ignored anyway. I still don't love this constructor but I think it accomplishes the desired goal.

src/viz/test/CMakeLists.txt
clevelam: All files are unique based on the number of processors.

src/viz/Ensight_Stream.hh
clevelam: Done.

src/viz/test/CMakeLists.txt
clevelam: Yap, it doesn't expressly test these in scalar mode anymore (because I made them parallel tests). The scalar versions of the test messed with python diff scripts. If we really need the scalar tests I can try and sort them out.

src/viz/Ensight_Translator.i.hh
clevelam: This worked great.

src/viz/test/CMakeLists.txt
clevelam: Actually there might be read contention for the serial golds (which happen for every processor count variation) but no write contention. I don't think this should be a real issue (because only a few processor will ever be run at the same time).

keadyk:
@clevelam Is it okay with you if I merge this once the GitHub checks are done? :)

clevelam:
> @clevelam Is it okay with you if I merge this once the GitHub checks are done? :)

Yes please! Thanks!

PR #869, opened 2020-07-03
Provide clang-tidy recommended changes

KineticTheory:
### Background

* Process four additional components with _clang-tidy_ code transformation recommendations and add these packages to the nightly regression that processes _clang-tidy_.

### Description of changes

* Provide updates for mesh_element, ode, linear, and memory components.
* Transform `typedef` to `using` commands.
* Make use of `auto` when the data type is specified on the RHS with a `static_cast`.
* Use `<cmath>` instead of `<math.h>`
* Use `std::array` instead of C-style arrays. 
  * https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-stack
  * https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#slcon1-prefer-using-stl-array-or-vector-instead-of-a-c-array
* When possible, use C++14 style range-based for-loops.
* Big changes in mesh_element take advantage of new C++14 capabilities to reduce the amount of code.

### Status

* Reference:
  * [Pre-Merge Code Review](https://github.com/lanl/Draco/wiki/Style-Guide)
  * [CDash Status](https://rtt.lanl.gov/cdash/index.php?project=Draco&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=-pr)
* Checks
  * [x] Travis/Appveyor CI checks pass
  * [x] Code coverage does not decrease
  * [x] [Valgrind test passes](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [x] [Toss3 checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [x] [Trinitite checks pass](https://rtt.lanl.gov/CDash/index.php?project=Draco)
  * [ ] Code reviewed/approved, sufficient DbC checks, testing, documentation

PR #871, opened 2020-07-10
Apply clang-tidy recommendations to more components

### LINKED ISSUE #874: gfortran-10.1.0 doesn't like z'00112233' used in drel.f90
Stop using hexadecimal initialization when not absolutely required.

KineticTheory:
### Background

* I am still trying to prepare Draco for a full CI check with clang-tidy.  See #869 for related changes.  Once these rules are working, we can expand the checks to include new patterns.
* Also fixes #874

### Description of changes

+ Use `std::array` instead of C-style arrays.
+ Use `nullptr` instead of `NULL`.
+ Replace `typedef` with `using`.
+ Use `auto` if the type is already provided by `static_cast`.
+ Use `std::move` for constructor arguments.
+ Use `= default` or `= delete` for trivial constructors.
+ Remove redundant `DLL_PUBLIC` decoration
+ Use `override` instead of `virtual` for overloaded member functions.
+ Use class member default initialization instead of constructor initialization.
+ Prune dead code blocks.
+ Make use of `emplace_back` for construct and append operations.

### Status

* Reference:
* Checks
* WIP
  * [x] cdi_eospac tests are failing on CTS-1


KineticTheory:
@attom @keadyk  If you have a few minutes, this PR could use a review.  The number of touched files is somewhat large.  Maybe you can pick a couple of directories to look at and leave the rest to someone else.

src/RTT_Format_Reader/RTT_Format_Reader.cc
```diff
@@ -17,12 +17,12 @@ namespace rtt_RTT_Format_Reader {
...
  * \param RTT_File Mesh file name.
  */
-RTT_Format_Reader::RTT_Format_Reader(string const &RTT_File)
+RTT_Format_Reader::RTT_Format_Reader(string RTT_File)
     : header(), dims(), spNodeFlags(), spSideFlags(), spCellFlags(),
       spNodeDataIds(), spSideDataIds(), spCellDataIds(), spCellDefs(),
       spNodes(), spSides(), spCells(), spNodeData(), spSideData(),
       spCellData() {
-  readMesh(RTT_File);
+  readMesh(std::move(RTT_File));
```
attom, line 25: I'm not up on the intricacies of moves and strings. This is a copy followed by a move?

src/cdi/CDI.cc
```diff
@@ -885,7 +886,7 @@ CDI::SP_MultigroupOpacity CDI::mg(rtt_cdi::Model m, rtt_cdi::Reaction r) const {
  */
 CDI::SP_CPEloss CDI::eloss(rtt_cdi::CPModelAngleCutoff mAC, int32_t pz,
                            int32_t tz) const {
-  map_it entry = CPEloss_map.find(std::make_pair(pz, tz));
+  auto const entry = CPEloss_map.find(std::make_pair(pz, tz));
```
attom, line 889: Why is auto preferred here?

src/RTT_Format_Reader/RTT_Format_Reader.cc
KineticTheory: That is correct. In this scenario, a copy must be made and the recommendation is to force the copy in the constructor function call reducing the number of steps required.  Gritty details are [here](https://releases.llvm.org/8.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-pass-by-value.html)

src/cdi_ipcress/IpcressDataTable.cc
```diff
@@ -107,13 +107,6 @@ void IpcressDataTable::setIpcressDataTypeKey() const {
         ipcressDataTypeKey = "ragray";
         dataDescriptor = "Gray Rosseland Absorption";
         break;
-      // NOTE: THIS KEY DOES NOT ACTUALLY EVER EXIST, See LA-UR-01-5543
```
attom, line 110: Do we want to keep a reference to this LA-UR somewhere? Something like `// To see what keys exist, see LA-UR-01-5543, "TITLE".`

src/cdi/CDI.cc
KineticTheory: The return type is fully defined by the type of `CPEloss_map`. Choosing the wrong type on the LHS will force a cast and possibly a deep object construction.

attom:
Otherwise, LGTM

src/cdi_ipcress/IpcressDataTable.cc
KineticTheory: Good idea.  I'll add the full reference to the class header.

KineticTheory: Fixed in 5a93fd0.

keadyk:
I'm looking at this now, @KineticTheory !

src/norms/Norms_Base.cc
```diff
@@ -15,14 +15,9 @@ using namespace rtt_norms;
...
-Norms_Base::Norms_Base(void)
+Norms_Base::Norms_Base()
     : d_sum_L1(-42.0), d_sum_L2(-42.0), d_Linf(-42.0), d_sum_weights(-42.0) {
-  reset();
-}
-
-//----------------------------------------------------------------------------//
-//! Destructor.
-Norms_Base::~Norms_Base() { /* empty */
+  Norms_Base::reset();
```
keadyk, line 20: I'm curious about this -- why is this reset() call prefixed with the class?

src/meshReaders/Hex_Mesh_Reader.hh
```diff
@@ -140,37 +140,32 @@ private:
  * <h3> Format Details </h3>
```
keadyk, line 140: It looks like most of the HTML (or whatever this is) was stripped out in the following lines, but not the <h3></h3> here -- why is that? Is this a clang-tidy oversight?

src/meshReaders/Hex_Mesh_Reader.cc
```diff
@@ -217,23 +219,19 @@ Hex_Mesh_Reader::get_unique_element_types() const {
...
+ *   faces.
  */
 std::map<std::string, std::set<unsigned>>
 Hex_Mesh_Reader::get_element_sets() const {
   // Alternatively, the private data of the class could be changed so that the
-  // work done here is done in the constructor. This would be more efficient
-  // if this is going to be used repetively.
-  typedef std::map<std::string, std::set<unsigned>> resultT;
+  // work done here is done in the constructor. This would be more efficient if
+  // this is going to be used repetively.
```
keadyk, line 233: This might be outside scope, but "repetitively" is misspelled :)

keadyk:
I have a bit of a hard time parsing the codecov/project results... is it saying this PR causes coverage of CDI to decrease? I'm not sure that makes sense to me.

src/norms/Norms_Base.cc
KineticTheory: That is an old code style.  Circa the 98 standard, there was a convention to use the base class name to scope calls like this.  Another one that is used in some codes is `this->reset()`.  It also ensures that you are not calling a `reset` function from a derived class.

src/meshReaders/Hex_Mesh_Reader.cc
KineticTheory: You've seem my other PRs; I'm all about fixing spelling errors.  MSVC has a built in spell checker that works like Word.  It underlines the misspelled words.

KineticTheory:
> I have a bit of a hard time parsing the codecov/project results... is it saying this PR causes coverage of CDI to decrease? I'm not sure that makes sense to me.

The number of covered lines is constant, but the total number of lines of code is less (I removed 14 lines of code in this PR), that means that coverage percent decreases.  I usually only worry about covering new lines of code.

src/meshReaders/Hex_Mesh_Reader.hh
KineticTheory: I just pushed a fix for this.  Again - old code.

KineticTheory:
d0e0e9f only changed comment blocks.  I'm not going to wait for the tests to run.  Merging now.

src/meshReaders/Hex_Mesh_Reader.hh
keadyk: (LOL, should've realized it would treat that as actual HTML :D )

PR #882, opened 2020-07-18
Apply clang-tidy transformations to parser component.

KineticTheory:
### Background

* More clang-tidy updates.

### Description of changes

+ Use `auto` when type is fully specified by RHS.
+ Use `strncpy` instead of `strcpy`.
+ Prefer `using` to `typedef`.
+ Remvove redundant `DLL_PUBLIC` decoration.
+ Use `override` for member functions of derived types.
+ Use `make_shared` instead of `reset` for shared pointers.
+ Use `=default` and `=delete` for trivial or disabled constructors.
+ Reindent and rewrap some comment blocks
+ Replace C-style arrays with `std::array`.
+ Use default initialization for some class member data.
+ Prefer C++ headers to C headers (`<cctype>` vs. `"cytpe.h"`).
+ Use `nullptr` instead of `NULL`
+ Eliminate dead-stores.
+ Make use of `FAIL_IF` in some places.
+ Make use of C++11 string literals to reduce inline escape characters.

### Status

* Reference:
* Checks

### DIFF
--- a/src/parser/Parse_Table.hh
+++ b/src/parser/Parse_Table.hh
@@ -15,7 +15,7 @@
 #include <vector>
 
 namespace rtt_parser {
-//-------------------------------------------------------------------------//
+//----------------------------------------------------------------------------//
 /*!
  * \brief Structure to describe a parser keyword.
  *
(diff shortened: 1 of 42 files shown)


src/parser/Parse_Table.hh
```diff
@@ -107,60 +101,53 @@ struct Keyword {
...
- * processing.
+ * For parsers designed to receive input from the console, the default behavior
+ * just described is not entirely appropriate. Console parsers are supported by
+ * treating any OTHER token whose text character is a semicolon as an empty
+ * keyword.  By default, semicolons are treated as whitespace and the parser
+ * will never see such a token.  A console stream can choose to convert an
+ * endline or other terminator to a semicolon token to force processing.
  */
-
-class DLL_PUBLIC_parser Parse_Table
```
attom, line 161: Why are we getting rid of DLL_PUBLIC_parser? No longer needed?

attom:
Otherwise, LGTM

src/parser/Parse_Table.hh
attom: (There are a few places where this is either removed or added)

KineticTheory: `DLL_PUBLIC_*` macros are no longer needed for classes and functions because recent versions of CMake automatically export those symbols for us.  

CMake doesn't do this for global data types (static class data or global scope data).  If such data is meant to be accessible via the shared object library, then these variables must still be marked as `DLL_PUBLIC_*`.

KineticTheory:
@attom Thanks for reviewing this. You may have noticed a few new warnings at https://rtt.lanl.gov/cdash/index.php?project=Draco for the _clang-tidy_ regressions.  I get those cleanup soon.  It is a bit unsettling that different versions of clang-tidy flag different things.

PR #898, opened 2020-09-02
Update cdi_ndi to fetch NDI_GENDIR_PATH at runtime

KineticTheory:
### Background

+ Previously, we required `NDI_GENDIR_PATH` to be set in the environment and discovered at configure time.  In practice, users will want this location to update automatically based on the current run environment (possibly after a `module swap` command). Making this change also reduces complicated build requirements (e.g.: spack) by deferring the choice of data set to run-time.
+ This run-time choice model was recommended by Saller, Dahl, and Grunau.

### Purpose of Pull Request

* [Fixes Redmine Issue #2048](https://rtt.lanl.gov/redmine/issues/2048)

### Description of changes

+ Remove logic from `CMakeLists.txt` used to discover `NDI_GENDIR_PATH`.
+ No longer save `NDI_GENDIR_PATH` to `config.h`.
+ Update constructors and tests in `cdi_ndi` to rely on ds++'s `Query_Env.hh` capabilities to read `NDI_GENDIR_PATH` at run time.
+ Add an informational message to tests when `NDI_GENDER_PATH` isn't found in  the test environment.  In this case, some tests are skipped.

### Status

* Reference:
* Checks

### DIFF
--- a/src/cdi_ndi/NDI_Base.cc
+++ b/src/cdi_ndi/NDI_Base.cc
@@ -17,7 +17,7 @@
 namespace rtt_cdi_ndi {
 
 //----------------------------------------------------------------------------//
-// File local helper functions
+// Auxillary functions
 //----------------------------------------------------------------------------//
 
 #ifdef NDI_FOUND
@@ -25,10 +25,10 @@ namespace rtt_cdi_ndi {
  * \brief Warn if NDI library version doesn't match GENDIR version to 2 digits.
  *
  * \param[in] gendir path to the gendir file
- * 
- * This function is a no-op if \c NDI_FOUND is false. 
+ *
+ * This function is a no-op if \c NDI_FOUND is false.
  */
-void warn_ndi_version_mismatch(std::string const &gendir) {
+void NDI_Base::warn_ndi_version_mismatch(std::string const &gendir) {
 
   std::string gendir_ver = rtt_dsxx::extract_version(gendir, 2);
   std::string ndi_ver = rtt_dsxx::extract_version(NDI_ROOT_DIR, 2);
@@ -50,7 +50,7 @@ void warn_ndi_version_mismatch(std::string const &gendir) {
 /*!
  * \brief Warn if NDI library version doesn't match GENDIR version to 2 digits.
  *        No-op when NDI_FOUND is false. */
-void warn_ndi_version_mismatch(std::string const & /*gendir*/) {}
+void NDI_Base::warn_ndi_version_mismatch(std::string const & /*gendir*/) {}
 #endif
 
 //----------------------------------------------------------------------------//
--- a/src/cdi_ndi/NDI_Base.cc
+++ b/src/cdi_ndi/NDI_Base.cc
@@ -9,10 +9,49 @@
 //----------------------------------------------------------------------------//
 
 #include "NDI_Base.hh"
+#include "ds++/DracoStrings.hh"
+#include "ds++/Query_Env.hh"
 #include "ds++/SystemCall.hh"
 #include "ds++/dbc.hh"
 
 namespace rtt_cdi_ndi {
+
+//----------------------------------------------------------------------------//
+// File local helper functions
+//----------------------------------------------------------------------------//
+
+#ifdef NDI_FOUND
+/*!
+ * \brief Warn if NDI library version doesn't match GENDIR version to 2 digits.
+ *
+ * \param[in] gendir path to the gendir file
+ * 
+ * This function is a no-op if \c NDI_FOUND is false. 
+ */
+void warn_ndi_version_mismatch(std::string const &gendir) {
+
+  std::string gendir_ver = rtt_dsxx::extract_version(gendir, 2);
+  std::string ndi_ver = rtt_dsxx::extract_version(NDI_ROOT_DIR, 2);
+  if (gendir_ver.size() == 0)
+    gendir_ver = "version unknown";
+  if (ndi_ver.size() == 0)
+    ndi_ver = "version unknown";
+  if (gendir_ver != ndi_ver) {
+    using DT = Term::DracoTerminal;
+    std::cout << "\n"
+              << Term::ccolor(DT::error) << "WARNING: In the cdi_ndi/NDI_Base "
+              << "constructor, the NDI library version (" << ndi_ver << ") is "
+              << "different than the NDI GENDIR version (" << gendir_ver
+              << "). \n"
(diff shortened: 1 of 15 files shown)


brryan:
@KineticTheory The recommendations of Saller, Dahl, and Grunau are good enough for me, of course, but does this approach not seem to risk having users end up using a Draco library compiled with a version of NDI that isn't compatible with the `gendir` file of the NDI version loaded at runtime? I don't see a good way to directly compare the NDI versions of the compiled NDI library and the loaded gendir file, but do you think it would still be worth printing a warning if the compile-time NDI_ROOT_DIR and runtime NDI_GENDIR_PATH don't agree? Maybe this concern shouldn't be taken that seriously, or such failures would always result in a descriptive NDI error, but I'm not sure right now.

Apart from that, this looks good to me (and it compiles and runs for me as expected if and only if I have `NDI_GENDIR_PATH` set appropriately at runtime).

brryan:
@KineticTheory From our email discussion, maybe the thing to do is try to get the NDI gendir version at runtime from the directory structure, but if we can't parse a major.minor.tweak version from the folder names, just throw a more generic warning such as "We could  NDI version of your gendir file at `NDI_GENDIR_PATH` is"?

KineticTheory:
* Saving offline discussion to this PR's discussion now that @dahl12 and @tgsaller are members.

> For comparing library/gendir versions, what is the best way to get the major.minor.tweak version of the gendir file? Can we expect this to always be available from the name of the parent directory of that gendir file? I didn't see version info in the gendir file itself.

>> It can be a concern, but only if the gendir itself if bad (i.e., if it has certain typos). We've only seen that for one special type of gendir. It's actually more a matter of using certain functionality rather than NDI version, but the latest version of NDI is the one with the fixed special gendir.
>> A warning is probably a good idea, but it should be with the whole major.minor.tweak.
>> In short, I think it shouldn't be an issue, but a warning wouldn't hurt.

>>> In the PR below, I modified Draco to use whatever NDI_GENDIR_PATH is found at runtime.  This removes any consistency check that would ensure that the NDI data and the linked NDI libraries are the same version.  Is this a concern?  What do you think about Ben’s recommendation about printing a warning?  Should the warning only trigger if the major.minor versions differ? What is the compatibility threshold wrt NDI version numbering?

KineticTheory:
> @KineticTheory From our email discussion, maybe the thing to do is try to get the NDI gendir version at runtime from the directory structure, but if we can't parse a major.minor.tweak version from the folder names, just throw a more generic warning such as "We could not determine what the NDI version of your gendir file at `NDI_GENDIR_PATH` is"?

I added the requested check and warning message.  Here is an example:

ndi_version_warning

KineticTheory:
@brryan Now I'm having trouble with the MSVC build.  I'll try to get this patched tonight.

KineticTheory:
> @brryan Now I'm having trouble with the MSVC build. I'll try to get this patched tonight.

I had to move a few things around to make MSVC happy.  I think I finally have a working version.

KineticTheory:
@brryan  All the tests are now clean.  Can you take one another look at this and merge it?

src/cdi_ndi/NDI_Base.cc
```diff
@@ -9,11 +9,50 @@
...
+/*!
+ * \brief Warn if NDI library version doesn't match GENDIR version to 2 digits.
+ *
+ * \param[in] gendir path to the gendir file
+ * 
+ * This function is a no-op if \c NDI_FOUND is false. 
+ */
+void warn_ndi_version_mismatch(std::string const &gendir) {
+
+  std::string gendir_ver = rtt_dsxx::extract_version(gendir, 2);
```
brryan, line 33: @KineticTheory Can `warn_ndi_version_mismatch` just be called in the `NDI_AtomicMass` constructor? It looks like duplicate code.

src/ds++/DracoStrings.hh
```diff
@@ -262,6 +262,39 @@ inline std::string remove_color(std::string const &colored_string) {
...
+ * \brief Extract version of the form M.m.p from a string.
+ *
+ * \param[in] string_in String that contains a version number as a substring.
+ * \param[in] digits number of version digits to print. Default = 3.
+ * \return The version as a substring.
+ *
+ * Given the string "/ccs/opt/vendors-ec/ndi/2.1.3/share/gendir.all", return
+ * "2.1.3".
+ */
+inline std::string extract_version(std::string const &string_in,
```
brryan, line 276: My understanding (I can't read regex though) is that this won't work for alpha/beta NDI deployments e.g. `2.1.3alpha/`. Maybe it's not worth agonizing about mismatches in those cases since they're alphas/betas, but just bringing it up now. An alternative would be to take a string as input, check that it is of the form "X.Y.Z", and then return three strings "X", "Y", "Z" to be compared with `std::string`'s `==` -- does that seem worthwhile to implement? If you think so I can make the modification.

brryan (commented):
@KineticTheory Looks good, and this latest commit works for me, just two comments I was curious for your opinion on before I hit merge. Thanks for doing this!

src/cdi_ndi/NDI_Base.cc
KineticTheory: Yeah.  I can make that change.  I didn't do it originally because I wanted this to be a file local function.  Give me a couple of minutes to fix this.

src/ds++/DracoStrings.hh
KineticTheory: I forgot about the `alpha/beta` versions.  Let me extend the regex to account for this case.  Thank you for the thoughtful review.

KineticTheory: I also need to add a unit test in `ds++` for this new function.

src/cdi_ndi/NDI_Base.cc
KineticTheory: Fixed.  Let's see what the CI reports.

src/ds++/DracoStrings.hh
KineticTheory: Also fixed.

brryan:
@KineticTheory Looks good to me, and CI looks good so far. Thanks for the changes and for adding the unit test. I can merge once the valgrind results come back, if you're happy with this.

KineticTheory:
> @KineticTheory Looks good to me, and CI looks good so far. Thanks for the changes and for adding the unit test. I can merge once the valgrind results come back, if you're happy with this.

The tests look good.  Merge when you are ready. Thanks!

PR #901, opened 2020-09-05
Fix a CrayPrg + spack configuration logic issue

### LINKED ISSUE #900: CMake can fail if SPACK_ROOT uses relative paths
@gaber reports

> Related to https://github.com/lanl/Draco/blob/develop/config/platform_checks.cmake#L81 :
> Spack seems to use a canonical path when it sets CC, CXX, etc. But the SPACK_ROOT environment variable may not be canonical. In other words, `CXX=/usr/projects/shavano/spack-environments/TRNI_19.0.4/1.4/spack/lib/spack/env/intel/icpc`
and `SPACK_ROOT=/usr/projects/shavano/spack-environments/modules/TRNI_19.0.4/../../TRNI_19.0.4/1.4/spack`
both refer to the same Spack installation, but the value of SPACK_ROOT doesn't appear in the value of CXX.

KineticTheory:
### Background

* A logic error related to discovery of Cray compile wrappers in `platform_checks.cmake` could lead to a configuration failure when building under spack.

### Purpose of Pull Request

* Fixes #900 

### Description of changes

* ~~Convert `SPACK_ROOT` to canonical form.~~
* Remove `query_craype` check as superfluous.
  * This check was helpful we we were still learning how to build and run on Cray machines, but isn't really that helpful now and gets in the way when building TRT codes via spack.

### Status

* Reference:
* Checks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,6 @@ dbsConfigInfo()
 # Platform Checks: Is HOST_NAME_MAX defined?  Is WinSock2.h available?  Is
 # gethostname() available?
 include( platform_checks )
-query_craype()
 set_draco_uname()
 query_have_gethostname()
 query_have_maxpathlen()
--- a/config/platform_checks.cmake
+++ b/config/platform_checks.cmake
@@ -64,48 +64,6 @@ Platform Checks...
 ")
 dbs_set_sitename()
 
-#--------------------------------------------------------------------------------------------------#
-# Sanity checks for Cray Programming Environments
-#
-# - Ensure CMAKE_EXE_LINKER_FLAGS contains "-dynamic"
-# - Ensure that the compilers given to cmake are actually Cray compiler
-#   wrappers.
-#--------------------------------------------------------------------------------------------------#
-macro( query_craype )
-
-  # We expect developers to use the Cray compiler wrappers. See also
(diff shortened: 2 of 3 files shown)


KineticTheory:
@gaber Does this change address your configuration failure?  Or do I need something more intrusive?  This block of code is to help our devs understand when their CrayPrg environment isn't correctly configured, but maybe this check is too much.

gaber:
> @gaber Does this change address your configuration failure? Or do I need something more intrusive? This block of code is to help our devs understand when their CrayPrg environment isn't correctly configured, but maybe this check is too much.

I think that might work, but I'm also not sure how Spack arrives at a canonical value for CXX. Does it use the equivalent of `realpath`? That would mean SPACK_ROOT could contain symlinks but CXX wouldn't, which might argue for using `REALPATH` instead of `ABSOLUTE`. But that wouldn't work if Spack didn't canonicalize CXX like that. And of course Spack's behavior could change over time...

It's also certainly possible that, when I ran into this, I had somehow constructed an inconsistent environment myself; i.e., that Spack normally arranges things so that SPACK_ROOT appears verbatim in the path to CXX, but that I managed to break that arrangement. But that brings me into agreement with your final comment; as you said, I also think the value of this check itself is sort of debatable.

RyanWollaeger (approved):
LGTM

PR #910, opened 2020-09-18
Fix the clang-tidy regression failure.

KineticTheory:
### Background

+ There was a bug in how we set compiler flags for clang that was causing the clang-tidy regression to fail with the error `clang-9: error: no input files`.   This was caused by an unexpected semicolon in the list of compiler flags.
+ Error shown on cdash for `clang-9.0.0-clangtidy_Debug-develop`.

### Purpose of Pull Request

* [Fixes Redmine Issue #2165](https://rtt.lanl.gov/redmine/issues/2165)

### Description of changes

+ Cleaning up the logic used to set compiler flags fixes this issue.
  - Use `string` instead of `list` to append compiler flags.
  - Deduplicate the list of compiler flags.
  - Ensure that flags set in the environment or as cmake options are captured.
+ Provide two new cmake functions to reduce cmake code duplication.
+ Process `tstMetis.cc` for 100-column formatting.

### Status

* Reference:
* Checks

### DIFF
--- a/src/VendorChecks/test/tstMetis.cc
+++ b/src/VendorChecks/test/tstMetis.cc
@@ -3,18 +3,16 @@
  * \file   VendorChecks/test/tstMetis.cc
  * \date   Wednesday, May 11, 2016, 12:01 pm
  * \brief  Attempt to link to libmetis and run a simple problem.
- * \note   Copyright (C) 2016-2019, Triad National Security, LLC.
- *         All rights reserved. */
+ * \note   Copyright (C) 2016-2020, Triad National Security, LLC., All rights reserved. */
 //------------------------------------------------------------------------------------------------//
 
+#include "metis.h"
 #include "ds++/Release.hh"
 #include "ds++/ScalarUnitTest.hh"
 #include <array>
-#include <metis.h>
 #include <vector>
 
-// Original provided by Erik Zenker
-// https://gist.github.com/erikzenker/c4dc42c8d5a8c1cd3e5a
+// Original provided by Erik Zenker, https://gist.github.com/erikzenker/c4dc42c8d5a8c1cd3e5a
 
 void test_metis(rtt_dsxx::UnitTest &ut) {
   idx_t nVertices = 10;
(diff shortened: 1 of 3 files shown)


src/VendorChecks/test/tstMetis.cc
```diff
@@ -3,18 +3,16 @@
  * \file   VendorChecks/test/tstMetis.cc
  * \date   Wednesday, May 11, 2016, 12:01 pm
  * \brief  Attempt to link to libmetis and run a simple problem.
- * \note   Copyright (C) 2016-2019, Triad National Security, LLC.
- *         All rights reserved. */
+ * \note   Copyright (C) 2016-2020, Triad National Security, LLC., All rights reserved. */
 //------------------------------------------------------------------------------------------------//
 
+#include "metis.h"
```
keadyk, line 9: Did clang tidy make this change? I thought the quote syntax was typically for directory local files -- is there some other use case I'm not aware of? :)

keadyk (approved):
This LGTM! I just had the one question about the Metis include :)

src/VendorChecks/test/tstMetis.cc
KineticTheory: No, I made that change.  

While there is no right-wrong answer to this question, the angle-brackets are typically reserved for system headers and are processed after all of the headers listed with quotes.  quotes should always be used for files in the current source directory.  TPL are neither local files or system files, so there is some ambiguity in how they should be treaded.  I made this change to so the Metis headers to be loaded before system headers.  I don't have a strong preference here.  If you want me to switch it back I can.

keadyk: Ahh, okay! I think it's fine this way. I was just curious. Your explanation makes sense, thanks!

keadyk:
@KineticTheory I will merge this shortly unless you object :)

KineticTheory:
> @KineticTheory I will merge this shortly unless you object :)

Merge away.  Thank you!

PR #930, opened 2020-10-15
WIP: New csk reader

attom:
This is a WIP and I anticipate needing several rounds of revision to get consistent with draco best-practices. My goal is to have this and downstream changes merged before the next code release.

### Background

* Currently, draco contains `src/compton/Compton.cc`, which is a wrapper around CSK functionality
* CSK as a library can read csk ASCII files and perform interpolation in temperature
* CSK also has exe to produce csk ASCII files

### Purpose of Pull Request

* CSK was built to be correct, not performant
* Reading ASCII files and interpolating are slow and memory-intensive in CSK
* This is the first step toward replacing Compton and the need to link to CSK
* Much like draco contains a reader/interpolator for opacity data, this adds functionality to read/interpolate Compton scattering data

### Description of changes

* Add preliminary reader/interpolator for Compton scattering data in `src/compton2`
* Compton2 class reads binary csk file and interpolates in temperature
* Compton2 is significantly faster than Compton (CSK)
* Added exe (CSKrw) to convert csk ASCII to binary, removing unneeded Legendre moments (for out_lin, out_nonlin, in_nonlin) and pre-computing temperature derivatives
* Added tests for CSKrw and Compton2
* This represents only a preliminary functionality; more will be added in the future
* For now, keep both Compton and Compton2

### Status

* Reference:
* Checks


KineticTheory:
@attom (1) I can help you with the Appveyor build issues. (2) Ignore the "Lint Code Base" failure.  (3) You should be able to fix the Travis failures for autodoc and style with relative ease.

I also want to help you clean up that long history so we don't see it in the merged 'develop' branch history.  I've been meaning to provide some guidance on this problem to other developers.  Let me work through the procedure write up and and I'll have you test it.

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+    std::cout << "\nCorrectly read group bounds and electron temps!" << std::endl;
+  }
+
+  // Test data retrieval: interpolate to a grid point in temperature
+  {
+    const double interp_T = T_evals[num_T_evals_gold - 1U];
+    std::cout << "Testing interpolation at T = " << (interp_T * mec2) << " keV\n";
+
+    std::vector<std::vector<std::vector<std::vector<double>>>> interp_data =
+        compton_test->interpolate_csk(interp_T);
```
KineticTheory, line 119: At some point, we should rethink this storage type.  This creates a lot of indirection.  (no change required for this PR.)

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+        2.906510852958e-23, 1.161410570085e-25, 3.652307822369e-25, -6.9213492314e-27, 0, 0, 0, 0,
+        6.471151078044e-25, -1.788886184269e-26, 9.067645193159e-25, 3.288019868804e-27,
+        // in_nonlin
+        2.412582001649e-20, 5.506360234252e-23, 1.858984933376e-22, -4.598103855448e-24, 0, 0, 0, 0,
+        1.70028840385e-22, -3.978450052949e-24, 4.964897275093e-22, 1.671777660171e-24,
+        9.356135881693e-24, -2.194753402066e-25, 0, 0, 0, 0, 9.680369923644e-24, -2.32267253788e-25,
+        2.875946793361e-23, 1.265594484598e-25, 4.274264612141e-25, -9.142968009692e-27, 0, 0, 0, 0,
+        5.392751048597e-25, -1.344066881751e-26, 8.87128629256e-25, 3.986007245261e-27};
+
+    ut.check(std::equal(flat_interp_data.begin(), flat_interp_data.end(), flat_interp_gold.begin(),
```
KineticTheory, line 182: simplify with `soft_equiv(a.begin(), a.end(), b.begin(), ...)`

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+      }
+    }
+
+    // Test sizes
+    ut.check(sz0 == num_evals_gold, "tested evals size");
+    ut.check((sz1 == num_groups_gold) && (sz2 == num_groups_gold), "tested groups size");
+    ut.check(sz3 == num_leg_moments_gold, "tested Legendre moments size");
+
+    // Print result (useful if golds need updating)
+    if (do_print) {
```
KineticTheory, line 220: Since you create/use this in more than one place, consider making this a helper function that is only defined at the top of this source file.

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+    compton_test.reset(new rtt_compton::Compton(filename));
+  } catch (rtt_dsxx::assertion &asrt) {
+    std::cout << "Draco exception thrown: " << asrt.what() << std::endl;
+    // We successfully caught the bad file!
+    caught = true;
+  } catch (const int & /*asrt*/) {
+    std::cout << "CSK exception thrown. " << std::endl;
+    // We successfully caught the bad file!
+    caught = true;
+  }
```
KineticTheory, line 295: You might also want the `catch(...)` case. You can omit it here and just let the top level catch work its magic.

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+  try {
+    // >>> UNIT TESTS
+    rtt_compton_test::compton_file_test(ut);
+    rtt_compton_test::compton_fail_test(ut);
+  }
+  UT_EPILOG(ut);
+}
+
+//------------------------------------------------------------------------------------------------//
+// End of test/tCompton.cc
```
KineticTheory, line 321: `t2Compton.cc`

src/compton2/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+#-----------------------------*-cmake-*----------------------------------------#
+# file   src/compton/CMakeLists.txt
+# author Andrew Till
```
KineticTheory, line 3: I think merging this into `compton` is best.  Let me know what you think.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
+//----------------------------------*-C++-*-----------------------------------//
+/*!
+ * \file   compton2/Compton2.cc
```
KineticTheory, line 3: As I mentioned earlier, let's find something more descriptive than `2`.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+ * \author Andrew Till
+ * \date   11 May 2020
+ * \brief  Implementation file for compton interface
+ * \note   Copyright (C) 2017-2020 Triad National Security, LLC.
+ *         All rights reserved. */
```
KineticTheory, line 8: Same comments as before wrt 100-col format.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+#include "compton2/Compton2.hh"
+#include "c4/C4_Functions.hh" //node()
+#include "c4/global.hh"
+#include "ds++/Assert.hh"
+
+#include <algorithm>
+#include <array>
+#include <cstring>
+#include <fstream>
+#include <iomanip> // RM
```
KineticTheory, line 21: Did you mean to remove some of these include directives?

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+// returns index in [0,len-2] so that xs[index] and xs[index+1] are valid
+// modifies x so that xs[index] <= x <= xs[index+1]
+UINT find_index(const std::vector<FP> &xs, FP &x) {
+  // Do binary search for index
+  auto loc = std::upper_bound(xs.begin(), xs.end(), x);
+  // loc - begin guaranteed to be in [0,len]
+  UINT index = (loc - xs.cbegin());
+
+  // Move index and x to be interior
+  // len >=2 assumed
```
KineticTheory, line 41: Add DbC to for `len >=2`.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+#include <ios>     // RM
+#include <iostream>
+
+using UINT = size_t;
+using FP = double;
+using vec = std::vector<FP>;
+
+namespace rtt_compton2 {
+
+// xs is assumed to be monotonically increasing and unique
```
KineticTheory, line 31: Please add doxygen style comment block.
Did you intend for this to be a file-scope free function?
Would it make more sense to define these as template functions `<UINT,FP>` and omit the `#define` above?

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  FP bL = (xR - x) / h;
+  FP bR = (x - xL) / h;
+
+  // Hermite functions
+  std::array<FP, 4> H{bL * bL * (3 - 2 * bL), bR * bR * (3 - 2 * bR), -h * bL * bL * (bL - 1),
+                      h * bR * bR * (bR - 1)};
+  return H;
+}
+
+Compton2::Compton2(std::string filename)
```
KineticTheory, line 64: Add Class doc block.

src/CMakeLists.txt
```diff
@@ -97,6 +97,7 @@ message(" ")
 message( STATUS "Configuring Level 5 packages --" )
 add_dir_if_exists( mesh )       # needs c4, ds++, mesh_element
 add_dir_if_exists( quadrature ) # needs mesh_element, parser, special_functions
```
KineticTheory, line 99: This PR needs to be checked with clang-tidy.  There are many simplifications that should be implemented, but it easier to iterate with the tool than for me to list them in this review.  I'll send instructions later.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+
+void Compton2::broadcast_MPI(int errcode) {
+  rtt_c4::global_max(errcode);
+  Insist(errcode == 0, "Non-zero errorcode. Exiting.");
+
+  int rank = rtt_c4::node();
+  constexpr int bcast_rank = 0;
+
+  // Broadcast sizes
+  UINT data_size = data_.size();
```
KineticTheory, line 88: This is confusing to me.  Maybe use `auto`?  This type will always be `size_t`, but `size_t` isn't always the same as `unsigned`.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  UINT p = 0;
+  num_temperatures_ = pack[p++];
+  num_groups_ = pack[p++];
+  num_leg_moments_ = pack[p++];
+  num_evals_ = pack[p++];
+  num_points_ = pack[p++];
+  data_size = pack[p++];
+
+  // Derived sizes
+  const auto tsz = int(num_temperatures_);
```
KineticTheory, line 101: Avoid c-style casts.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  const auto egsz = int(num_groups_ + 1U);
+  const auto fgsz = int(num_temperatures_ * num_groups_);
+  const auto isz = int(fgsz + 1U);
+  const auto dsz = int(data_size);
+
+  // Broadcast grids
+  if (rank != bcast_rank)
+    Ts_.resize(tsz);
+  rtt_c4::broadcast(&Ts_[0], tsz, 0);
```
KineticTheory, line 110: Avoid using `&vec[0]` without extra precautions as this is undefined behavior if the vector is empty.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  // Broadcast data itself
+  if (rank != bcast_rank)
+    data_.resize(dsz);
+  rtt_c4::broadcast(&data_[0], dsz, 0);
+  if (rank != bcast_rank)
+    derivs_.resize(dsz);
+  rtt_c4::broadcast(&derivs_[0], dsz, 0);
+}
+
+int Compton2::read_binary(std::string filename) {
```
KineticTheory, line 132: Pass strings by `const &` unless you intend to use `std::move`.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  // Read
+  auto fin = std::ifstream(filename, std::ios::in | std::ios::binary);
```
KineticTheory, line 134: Check for failure?

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+
+  // Ensure valid type
+  char expected[] = " csk ";
```
KineticTheory, line 137: Clang-tidy will require you to change some of this.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  char file_type[sizeof(expected)];
+  fin.read(file_type, sizeof(file_type));
+  if (std::strcmp(file_type, expected) != 0) {
+    std::cerr << "Expecting binary file " << filename << " to start with '" << expected
+              << "' but got '" << file_type << "'";
+    std::cerr << std::endl;
+    return 1;
+  }
+
+  UINT binary_ordering, version_major, version_minor;
```
KineticTheory, line 147: One line per declaration with default initialization.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  fin.read(reinterpret_cast<char *>(&version_major), sizeof(UINT));
+  fin.read(reinterpret_cast<char *>(&version_minor), sizeof(UINT));
+  fin.read(reinterpret_cast<char *>(&binary_ordering), sizeof(UINT));
+  if (version_major != 1 || binary_ordering > 1) {
```
KineticTheory, line 151: Check for read errors or state of the istream?
It would likely be more efficient to read the whole file into a buffer and then parse the buffer?

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+  num_temperatures_ = tsz;
+  num_groups_ = gsz;
+  num_leg_moments_ = lsz;
+  num_evals_ = esz;
+  // point is (leg moment, eval) pair
+  // first eval has all leg moments; others have only the 0th moment
+  num_points_ = num_evals_ + num_leg_moments_ - 1U;
+  UINT egsz = gsz + 1;
+
+  if (false) {
```
KineticTheory, line 181: Remove or add a comment.  Consider `#if 0` to keep this out of the binary.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+    std::cout << std::defaultfloat;
+
+    std::cout << "-------------------------------------------------\n";
+    std::cout << '\n';
+  }
+
+  return 0;
+}
+
+////////////////////////////////////////////////////////////////////////////////
```
KineticTheory, line 266: Please use standard draco formatting.

src/compton2/Compton2.cc
```diff
@@ -0,0 +1,475 @@
...
+void Compton2::interp_dense_inscat(vec &inscat, double Te_keV, size_t num_moments_truncate) const {
+  // Ordering of inscat is 1D array (slow) [moment, group-to, group-from] (fast)
+
+  // Finds index and nudges Teff st Ts_[index] <= Teff <= Ts_[index+1]
+  // and 0 <= index <= Ts_.size()-2;
+  FP Teff = Te_keV;
+  UINT iT = rtt_compton2::find_index(Ts_, Teff);
+
+  // Fill Hermite function
+  std::array<FP, 4> H = rtt_compton2::hermite(Teff, Ts_[iT], Ts_[iT + 1U]);
```
KineticTheory, line 277: Avoid the use of single letter variables except for look indices.  It is difficult to search for such variables.

src/compton2/Compton2.hh
```diff
@@ -0,0 +1,197 @@
+//----------------------------------*-C++-*-----------------------------------//
+/*!
```
KineticTheory, line 2: Use standard 100-col format (I know - we adopted this format after you started this branch. sorry.)

src/compton2/Compton2.hh
```diff
@@ -0,0 +1,197 @@
...
+//----------------------------------------------------------------------------//
+
+#ifndef rtt_compton2_Compton2_hh
+#define rtt_compton2_Compton2_hh
+
+// C++ standard library dependencies
+#include <string>
+#include <vector>
+// Draco libraries
+#include "compton2/Sparse_Compton_Matrix.hh"
```
KineticTheory, line 18: Remove this comment and let clang-format order these include directives.  The stdlib headers should be listed after draco headers.

src/compton2/Compton2.hh
```diff
@@ -0,0 +1,197 @@
...
+
+namespace rtt_compton2 {
+
+enum Eval : size_t { in_lin = 0, out_nonlin = 1, nl_diff = 2 };
```
KineticTheory, line 22: prefer enum class.

KineticTheory (changes requested):
This isn't a full review, but I think it will get you started on some of the changes I'm requesting.  I'll update my notes after a 2nd pass.

keadyk:
Woot!! Nice, @attom! Let me know if you want more eyes on this (I don't want to jump the gun if you're still working on it :) )

src/CMakeLists.txt
```diff
@@ -97,6 +97,7 @@ message(" ")
...
+add_dir_if_exists( compton2 )     # needs ds++, units, cdi_analtyic
```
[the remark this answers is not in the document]
attom: I like the idea of renaming both, though my initial goal was to not break anything that depended on `compton`. If I rename the directory, will CMakeLists.txt in CTA/TRT/Jayenne be smart enough to know where to look?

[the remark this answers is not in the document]
attom: Thoughts on what to name the directories? I like the idea of having them separate, because they do very different things. `compton` wraps an interface to `csk` while `compton2` handles reading / interpolating csk data files itself.

`compton` -> `compton_wrapper`
`compton2` -> `compton_tools`

?

src/compton/test/CMakeLists.txt
```diff
@@ -45,6 +48,13 @@ else()
     LABEL   "nomemcheck"
     MPI_PLUS_OMP )
 
+  add_parallel_tests(
+    SOURCES "${test2_sources}"
+    PE_LIST "1"
+    DEPS    "Lib_compton;Lib_c4"
```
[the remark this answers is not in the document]
attom: Will try without

src/compton/test/CMakeLists.txt
```diff
@@ -16,6 +16,9 @@ project( compton_test CXX )
 set( test_sources
   ${PROJECT_SOURCE_DIR}/tCompton.cc )
 
+set( test2_sources
+  ${PROJECT_SOURCE_DIR}/t2Compton.cc )
```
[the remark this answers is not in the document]
attom: Agreed. Will update. This test uses dummy data, so maybe include that in the name?

src/compton/test/CMakeLists.txt
```diff
@@ -45,6 +48,13 @@ else()
     LABEL   "nomemcheck"
     MPI_PLUS_OMP )
 
+  add_parallel_tests(
+    SOURCES "${test2_sources}"
+    PE_LIST "1"
+    DEPS    "Lib_compton;Lib_c4"
+    LABEL   "nomemcheck"
```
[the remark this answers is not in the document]
attom: Not that I'm aware of. Probably a copy-paste error from wherever I got the CMake. Will try removing

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
+//-----------------------------------*-C++-*----------------------------------//
```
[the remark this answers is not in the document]
attom: I'll look around for one

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+/*!
+ * \file   compton/test/t2Compton.cc
+ * \author Andrew Till
+ * \date   2020 Oct 14
+ * \brief  Implementation file for t2Compton
+ * \note   Copyright (C) 2017-2020 Triad National Security, LLC.
```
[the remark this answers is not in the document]
attom: Will update

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+  const double tol = 1e-11;
+
+  // Start the test.
+
+  std::cout << "\n---------------------------------------------------------\n"
+            << "   Test Draco code calling CSK_generator routines\n"
+            << "---------------------------------------------------------\n";
+
+  // open a small mg opacity file:
+  const std::string filename = ut.getTestSourcePath() + "../../compton2/test/dummy_data";
```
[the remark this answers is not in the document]
attom: I wanted to be sure that compton and compton2 were getting the same answer for the same input, which is why I wrote this test. They are. Since compton is already tested, should I just remove this test? It's using the data in compton2 but the methods in compton.

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+  std::cout << "Attempting to construct a Compton object...\n" << std::endl;
+  std::unique_ptr<rtt_compton::Compton> compton_test;
+
+  try {
+    compton_test.reset(new rtt_compton::Compton(filename));
+  } catch (int /*asrt*/) {
+    FAILMSG("Failed to construct a Compton object!");
+    // if construction fails, there is no reason to continue testing...
+    return;
+  }
```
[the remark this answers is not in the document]
attom: Can do. This is a copy-paste from `src/compton/test/tCompton.cc`

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+  std::cout << "\n(...Success!)" << std::endl;
+
+  // Check some of the data in the CSK_generator-opened file:
+  const std::vector<double> grp_bds = compton_test->get_group_bounds();
+  const std::vector<double> T_evals = compton_test->get_etemp_pts();
+
+  // Unitless (divided by mec2)
+  // NB: These values can be read directly from the ASCII data files
+  // (3rd line)
```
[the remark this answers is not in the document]
attom: sorry, 3rd line from the csk data file. I'll edit the comment to be more explicit.

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+  /*
+  for (size_t i = 0; i < T_evals_gold.size(); ++i)
+  {
+      T_evals_gold[i] *= mec2;
+  }
+  line2_Ts_gold[0] *= mec2;
+  line2_Ts_gold[1] *= mec2;
+  */
+
+  Ensure(grp_bds.size() == (num_groups_gold + 1U));
```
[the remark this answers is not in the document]
attom: Yeah, another copy-paste from `src/compton/test`. I'll fix.

src/CMakeLists.txt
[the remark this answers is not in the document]
warsa: How about csk_interface or compton_interface for the "wrapper" part, compton_data for the data "reading/handling" part?

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+  ut.check(std::equal(grp_bds.begin(), grp_bds.end(), grp_bds_gold.begin(),
```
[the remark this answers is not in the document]
attom: Oh, nifty! Is that `soft_equiv_deep`?

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+                      [tol](double a, double b) -> bool { return soft_equiv(a, b, tol); }),
+           "checked group boundaries");
+
+  Ensure(T_evals.size() == num_T_evals_gold);
```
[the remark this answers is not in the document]
attom: Will fix.

src/compton/test/t2Compton.cc
```diff
@@ -0,0 +1,322 @@
...
+  ut.check(std::equal(T_evals.begin(), T_evals.end(), T_evals_gold.begin(),
+                      [tol](double a, double b) -> bool { return soft_equiv(a, b, tol); }),
+           "checked temperature grid");
+
+  if (!soft_equiv(compton_test->get_min_etemp(), line2_Ts_gold[0]))
+    FAILMSG("Min etemp read incorrectly!");
```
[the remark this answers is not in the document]
attom: I wish I knew the Draco testing macros / functions better. Maybe we should have a review meeting for Draco developers?

src/compton/test/t2Compton.cc
attom: Oh yeah. This exact point is a major reason why we're moving from compton (which uses vector<vector<vector<vector<double>>>>) to compton2 (which uses vector<double>)

src/compton/test/t2Compton.cc
attom: what's the last argument? the tolerance?

src/compton/test/t2Compton.cc
attom: will do

src/compton/test/t2Compton.cc
attom: Is it just replace the catch statement with " catch(...)" or does something special need to replace the ellipses. I'm not overly familiar with this language feature or exceptions in general

src/compton/test/t2Compton.cc
attom: good catch

src/compton2/CMakeLists.txt
attom: I'd like to keep them separate so that we could remove the entire `compton` directory later if we wanted. Otherwise, we'd have to be very careful about what was a standalone functionality and what was a wrapper.

src/compton2/Compton2.cc
attom: yep

src/compton2/Compton2.cc
attom: will update

src/compton2/Compton2.cc
attom: yeah, they were used while I was debugging. good catch

src/compton2/Compton2.cc
attom: good call

src/compton2/Compton2.cc
attom: I'll dig around for one. Yes, the function is a helper function that doesn't need to be in the class, though I'm not sure who else would use it.

We could templatize on UINT and FP. Were you thinking just for the free functions or for the entire class?

src/compton2/Compton2.cc
attom: I'll dig around for one. Is anyone using Doxygen these days?

src/CMakeLists.txt
attom: Thanks! Should I do all these edits on the squashed branch or the current branch?

src/compton2/Compton2.cc
attom: I was hoping to get your help more generally on the broadcast / receive and read functions to be sure they're architecture (Endian-ness, etc.) portable.

Perhaps we could have a WebEx on what to do for these?

src/compton2/Compton2.cc
attom: Do you prefer static_cast<int>?

src/compton2/Compton2.cc
attom: See above comment. Not sure the best way to do the broadcast for a vector.

src/compton2/Compton2.cc
attom: Good best-practices. Will do.

src/compton2/Compton2.cc
attom: Do I use `fin.good()` for that? And throw if not good?

src/compton2/Compton2.cc
attom: The binary file is a set of chars that starts with " csk ". How should I compare to the start of the file if not char[]?

src/compton2/Compton2.cc
attom: sure

src/compton2/Compton2.cc
attom: I defer to you. Should one rank read the file, put it in a buffer, and then MPI communicate that buffer to the other ranks and have everyone parse?

src/compton2/Compton2.cc
attom: Yeah, this is a debug print statement. Do you think there's value in making a function that prints the contents of a file or should I just remove this (or keep around with `#if 0`)?

src/compton2/Compton2.cc
attom: Must have been a copy-paste from somewhere. Is standard draco formatting `/* ---- ... ---- */`?

src/compton2/Compton2.cc
attom: Will change to `hermite_eval` or similar. Preferences on camel case vs underscore?

src/compton2/Compton2.hh
attom: should be easy enough to find a place to copy-paste from

src/compton2/Compton2.hh
attom: I'll check the squashed version to see what git-format did

src/compton2/Compton2.hh
attom: will change

attom:
Howdy @keadyk. Yes, another pair of eyes would be most appreciated, but perhaps after my next commit, since there's a lot to be done. While I have you, what do you think about naming? @KineticTheory and I are debating how to name the directories, currently the unhelpful `compton` and `compton2`, or whether they should be merged into one directory, which I think would be more confusing than helpful. Thoughts?

src/CMakeLists.txt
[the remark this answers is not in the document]
KineticTheory: > I like the idea of renaming both, though my initial goal was to not break anything that depended on `compton`. If I rename the directory, will CMakeLists.txt in CTA/TRT/Jayenne be smart enough to know where to look?

Draco's build system publishes all of the path and library information automatically.  In cta, you will need to tell the code to link the libraries `compton_interface` and `csk_interface` (or whatever names we end up with).  Any C++ source files that currently has `#include compton/Compton.hh` will need to be updated.  This can be done easily with a sed command or through an IDE _refactor_ command.

src/compton/test/CMakeLists.txt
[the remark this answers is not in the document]
KineticTheory: > Agreed. Will update. This test uses dummy data, so maybe include that in the name?

That is the pattern I've used for some of the `cdi_*` tests, but I'm not too picky about names in test directories.

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
KineticTheory: Look in `ds++`.  This is a new change in Draco and we are still updating some of the files that didn't get fixed by my one-pass sed script.  If you use Draco's Emacs helpers, the updated dividers are available from within Emacs via key bindings.   `M-x draco-mode`, followed by `M-x describe-mode` and scrolling down to the section on 'draco minor mode'.

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
KineticTheory: You can leave it here.  I just wanted to make sure that this design was intentional.

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
KineticTheory: > Can do. This is a copy-paste from `src/compton/test/tCompton.cc`

Again, this can be left here if you are okay with the test aborting immediately if some other exception is thrown.

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
KineticTheory: > Oh, nifty! Is that `soft_equiv_deep`?

That is the actual implementation. There are specializations for 1, 2 and 3D `std::vector` (or you can add your own).  See https://github.com/lanl/Draco/blob/dd14663b05f77866335596b3c52d39fdd8964a1b/src/ds%2B%2B/Soft_Equivalence.hh#L204 and https://github.com/lanl/Draco/blob/dd14663b05f77866335596b3c52d39fdd8964a1b/src/ds%2B%2B/test/tstSoft_Equiv.cc#L195

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
KineticTheory: > I wish I knew the Draco testing macros / functions better. Maybe we should have a review meeting for Draco developers?

That is probably a good thing to do.  I just takes a lot of energy that I seem to be in short of lately. 😄

src/compton/test/t2Compton.cc
KineticTheory: > what's the last argument? the tolerance?

Yes. 

```c++
template <typename Value_Iterator, typename Ref_Iterator>
inline bool
soft_equiv(Value_Iterator value, Value_Iterator value_end, Ref_Iterator ref, Ref_Iterator ref_end,
           typename std::iterator_traits<Value_Iterator>::value_type const precision = 1.0e-12);
```

or 

```c++
inline bool soft_equiv(const std::vector<double> &value, const std::vector<double> &ref,
                       const double precision = 1.0e-12
```

src/compton/test/t2Compton.cc
KineticTheory: > Is it just replace the catch statement with " catch(...)" or does something special need to replace the ellipses. I'm not overly familiar with this language feature or exceptions in general

Since this is is a test, I think `catch(...)` is appropriate.  You can fine tune the error message or recovery behavior with a more detailed `catch(arg)`, if there is need to do so.

src/compton2/Compton2.cc
KineticTheory: > I'll dig around for one. Yes, the function is a helper function that doesn't need to be in the class, though I'm not sure who else would use it.
> 
> We could templatize on UINT and FP. Were you thinking just for the free functions or for the entire class?

I was thinking for the entire class since you use these CPP macros in a way that a native C++ template could be used.  The template form is less error prone and the static analysis tools do a better job of checking.

src/compton2/Compton2.cc
KineticTheory: * Try using one from `ds++/DracoStrings.hh` 
* Something like:
```c++
/*!
 * \brief one-liner
 *
 * \param[in] string_in This string will be converted letter by letter to lowercase.
 * \return A string that contains no uppercase letters.
 *
 * Longer description as needed, with design, intent, possibly examples.
 */
```

src/compton2/Compton2.cc
KineticTheory: Historically, our developers prefer the underscore separator.

KineticTheory:
@attom I hope I didn't discourage you with that long list of comments.  It is great to see these new capabilities going into the code base!  Let me know how else I can help.

src/compton2/Compton2.cc
KineticTheory: I really think that decision is up to you.  In more mature code, I would recommend that the code be removed entirely. Since this is brand new code, it might be worth keeping it in place locally.  Optionally, you could create a utility application that reads the file and dumps the contents to the screen.  We have something similar for both cdi_ipcress and cdi_eospac.

keadyk:
@attom I agree with you -- I think it would be hard to parse if everything were in the same directory. Since we will likely scrap the original compton folder in the future, were you thinking of just renaming compton2 -> compton at that point? :)

Hmmm. Maybe we could go with something like csk_compton and fast_compton? It seems like adding "_reader" to the end of the directory name would be too unwieldy, and not entirely accurate. *ponders*

attom:
What do you think of `compton_wrapper` (for `compton`) and `compton_tools` (for `compton2`)? Or `compton` -> `csk_wrapper`? Or `compton` -> `csk_compton` as you suggest. The difficulty is I've been using csk for both the library and the file format.

keadyk:
Ohh, okay, I see! Let me think about this a little more. My first impression is that I think I like the "compton_tools" name for your new code. :)

keadyk:
@attom Ok, I think the compton_wrapper / compton_tools naming scheme makes the most sense to me. I also use "csk" to refer to the library, the data, the actual expression for the kernel, etc., so that's probably not a helpful way to differentiate them (as you pointed out) :)

src/CMakeLists.txt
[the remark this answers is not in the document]
attom: Alright, time for a vote: 

`compton` -> `compton_interface`

or

`compton` -> `compton_wrapper`

@warsa or @keadyk or @KineticTheory ?

I'm planning on using `compton2` -> `compton_tools` unless there is an objection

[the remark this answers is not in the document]
attom: I'm planning on using `compton_interface` unless there is a major objection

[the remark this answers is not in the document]
keadyk: Sounds good to me! :)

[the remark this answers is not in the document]
warsa: I prefer it as well.

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
attom: Yes, that's fine with me

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
attom: Where do we want the ending `*/`? Same line or next line down or does it not matter?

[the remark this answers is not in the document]
KineticTheory: Prefer the same line as the copyright.

src/compton/test/t2Compton.cc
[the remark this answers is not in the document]
attom: Excellent. I'll have to add this to my other unit tests in consuming codes

src/compton2/Compton2.cc
attom: I need to think more about this. I'd like FP and UINT to be the same as the data types that are stored in the binary csk data file but then I'd like to return doubles and size_t or uint64_t's. Does that seem more reasonable?

KineticTheory: That makes perfect sense.  CSK should be able to publish the data types through a header or `config.h` or even through the cmake system. (`csk-config.cmake`).  You might even be able to use to C++'s [decltype](https://en.cppreference.com/w/cpp/language/decltype) to return the type of a data field defined by CSK.

```c++
class compton2 {
private:
  // DATA
  decltype(csk::fp_type) csk_data;
  double compton2_data;
...}
```

I'm not sure if that snippet is valid C++, but you get the idea.

attom:
Alright, I've addressed most of the comments above. Big things that I haven't addressed that will need more thought:

* FP and UINT: should we template on these? how do we divorce the underlying data format of the binary file from the interface to the class?
* broadcasting: best practices for broadcasting a vector?
* binary read: best practices for reading a vector of unsigned 64-bit ints or doubles from a character stream in a way that is correct for multiple architectures?
* comparing an array of chars from a binary file to a known string; should I use std::array<char, length> and std::equal?
* how to use clang-tidy on this
* I'll resubmit as a PR with a single commit once we're closer to merging

Thanks @KineticTheory for guiding me through this so far! If anyone else has concerns, I think we're getting close to a good time for y'all to voice them

KineticTheory:
> * binary read: best practices for reading a vector of unsigned 64-bit ints or doubles from a character stream in a way that is correct for multiple architectures?

This is straight forward and there are a ton of examples on line.  Let me know if you need any help.  I've done this a few times, but I don't think any of that code is in Draco.  The main thing to avoid for portability is **Fortran**.  Fortran does non-portable padding to the header/footer blocks with binary files.  I can't remember how _Endianess_ plays with binary data.  However, `ds++/Endian` provides some help if you need to swap data around.

> * comparing an array of chars from a binary file to a known string; should I use std::array<char, length> and std::equal?

* I would probably ready the data in binary and stuff it into a `std::string` and use simple `==`  checks.  When saving strings to a binary file, you normally save the field size and then the data.  That way you know how many bytes to load and stuff into the string container.  `ds++/DracoStrings` may have some useful functions to cast raw data into strings and other types.

> * how to use clang-tidy on this

```console
ccscs% module swap gcc llvm
ccscs% cmake C_FLAGS=-Werror CXX_FLAGS=-Werror cmake -DDRACO_STATIC_ANALYZER=clang-tidy ~/draco
ccscs% make -j 20 -k  # build as much as you can before running with VERBOSE=1
ccscs% make VERBOSE=1
```

* Then use http://clang.llvm.org/extra/clang-tidy/checks/list.html to understand why clang-tidy is complaining.

KineticTheory:
When you are looking at *clang-tidy*, let me know if there are additional rules that you think should be enabled by default.

attom:
Whew. Applied clang-tidy to both `compton_interface` and `compton_tools`. Surprisingly, fewer changes than I thought, which might just mean we need to add more checks to clang-tidy.

Tomorrow, I hope to look at portability (Endianness, etc.)

attom:
@KineticTheory, do we have a system with non-standard endianness for testing? Or a system that isn't 64 bit in the normal way?

KineticTheory:
> @KineticTheory, do we have a system with non-standard endianness for testing? Or a system that isn't 64 bit in the normal way?

Actually, I don't think we have any big-endian machines right now.  Sequoia was big-endian but that machine is gone.  Let's punt on this concern for now.

src/compton_tools/python/common_compton.py
```diff
@@ -1,22 +1,23 @@
 #!/usr/bin/env python
-#---------------------------------------*-python-*-------------------------------------------------#
+# --------------------------------------*-python-*------------------------------------------------ #
```
KineticTheory, line 2: Did the linter really want those extra spaces?  I guess I need to update the Emacs template to match.

attom:
Open questions:

* Read the whole file into a buffer and then parse the buffer vs current strategy of read-as-you-go (can we defer this for a future merge? I'm not quite sure the mechanics of the former option)

* Is broadcasting okay as it stands for now? I've added checks for sizes.

* FP and UINT; not sure what I want or what the best thing to do is. Goals are:
  * Have interface return doubles, since that's what the rest of our codes use for floating-point types
  * Read and write to the binary file using FP (currently double) and UINT (currently uint64_t) for consistency
  * Safely convert as needed from FP to doubles using static_cast<double>(FP val); I assume this is currently a no-op?
  * Question: Should the conversion happen during the read phase or the interpolate phase? (@KineticTheory or @keadyk, thoughts?)

keadyk:
@attom I would think it'd be easiest to convert when you read (if I understand correctly!), but I don't have a strong opinion :)

attom:
That works for me. I'll play around with the change tomorrow.

attom:
Alright, finally getting back around to this. @keadyk and @KineticTheory could you please take another look? Once we're happy, I'll close this PR and make a new branch with just one commit of the changes

keadyk:
I admittedly did some skimming, but what I've seen looks reasonable to me! Excited to see it in action :)

KineticTheory:
> Alright, finally getting back around to this. @keadyk and @KineticTheory could you please take another look? Once we're happy, I'll close this PR and make a new branch with just one commit of the changes

I should have some time this evening or tomorrow afternoon to give this another review.  Sorry for the delay.

src/compton_tools/Sparse_Compton_Matrix.hh
```diff
@@ -0,0 +1,101 @@
...
+namespace rtt_compton_tools {
+//------------------------------------------------------------------------------------------------//
+/*!
+ * \class Sparse_Compton_Matrix
+ *
+ * \brief Stores a Compton scattering matrix sparsely and allows
+ *        matrix multiplication; can store an arbitrary number of points,
+ *        where each point represents an  M x G x G matrix (moments, groups)
+ *
+ *        THIS CLASS IS A STUB (not yet fully implemented or used)
```
KineticTheory, line 25: If this isn't needed, then maybe is shouldn't be in the PR?  Or maybe you can leave it in a `#ifdef 0` block for now.

src/compton_tools/cskrw.cc
```diff
@@ -0,0 +1,1023 @@
...
+  // index with T and gfrom; applies to all points
+  std::vector<UINT> indexes;
+
+  // sparse version of Dense_Compton_Data's data and derivs
+  // [point, T, gfrom, gto]
+  std::vector<FP> data;
+  std::vector<FP> derivs;
+};
+
+Sparse_Compton_Data::Sparse_Compton_Data() : first_groups(), indexes(), data(), derivs() {}
```
KineticTheory, line 56: I think this can be removed.

src/compton_tools/cskrw.cc
```diff
@@ -0,0 +1,1023 @@
...
+                           classical_electron_radius * rtt_units::AVOGADRO;
+  Ensure(soft_equiv(csk_opac_norm, 2 * 0.037558, 1e-4));
+  //
+  // convert from opacity (cm^2/mole) to micro xs (cm^2)
+  //const FP csk_xs_conv = 1. / rtt_units::AVOGADRO;
+  //Ensure(soft_equiv(2 * 0.2003102 * csk_xs_conv, 6.6524587e-25, 1e-4));
+
+  // VALUES USED IN CSK'S PHYSICAL_CONSTANTS.HH
+  // Normalization for energy and nonlinear terms
+  const FP mec2 = 510.998;
```
KineticTheory, line 137: * Probably better to pull these values directly from csk.
* This file looks a bit rough, maybe just add some history and discussion to the top of the file.  It doesn't appear to be something that would be consumed by users.

src/compton_tools/python/common_compton.py
```diff
@@ -0,0 +1,89 @@
...
+# ------------------------------------------------------------------------------------------------ #
+# TPL
+import numpy as np
+# ------------------------------------------------------------------------------------------------ #
+
+# These are the common functions that are used to read and write Compton data
+
+
+# ------------------------------------------------------------------------------------------------ #
+def print_grids(grids, fileroot, verbosity=False):
```
KineticTheory, line 21: @alexrlongne Can you review the python here?

KineticTheory (changes requested):
This is looking much more mature.  Thanks for working on it. 

I still want to look at the code coverage numbers before approving.

.clang-format
```diff
@@ -61,7 +61,7 @@ ForEachMacros:
   - BOOST_FOREACH
 IncludeBlocks:   Preserve
 IncludeCategories:
-  - Regex:           '^"(c4|cdi|cdi_analytic|cdi_CPEloss|cdi_eospac|cdi_ipcress|cdi_ndi|compton|device|diagnostics|ds\+\+|experimental|fit|linear|memory|mesh|mesh_element|meshReaders|min|norms|ode|parser|plot2D|quadrature|rng|roots|RTT_Format_Reader|special_functions|timestep|units|viz)/'
+  - Regex:           '^"(c4|cdi|cdi_analytic|cdi_CPEloss|cdi_eospac|cdi_ipcress|cdi_ndi|compton_interface|compton_tools|device|diagnostics|ds\+\+|experimental|fit|linear|memory|mesh|mesh_element|meshReaders|min|norms|ode|parser|plot2D|quadrature|rng|roots|RTT_Format_Reader|special_functions|timestep|units|viz)/'
```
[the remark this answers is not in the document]
attom: Done (will appear in next commit)

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
+//--------------------------------------------*-C++-*---------------------------------------------//
+/*!
+ * \file   compton_tools/Compton_Native.cc
+ * \author Andrew Till
+ * \date   11 May 2020
+ * \brief  Implementation file for native compton bindary-read and temperature interpolation
```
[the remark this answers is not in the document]
attom: Fixed

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+//------------------------------------------------------------------------------------------------//
+// FREE-FLOATING HELPER FUNCTIONS
+//------------------------------------------------------------------------------------------------//
+
+//------------------------------------------------------------------------------------------------//
+/*!
+ * \brief find location in a sorted list and min/max value to be within list
+ *
+ * \param[in] xs vector of monotonically increasing and unique values
+ * \param[inout] x value whose location in xs needs to be found;
```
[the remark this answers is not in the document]
attom: Done

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+size_t find_index(const vec_d &xs, double &x) {
+  Require(xs.size() >= 2);
+
+  // Do binary search for index
+  auto loc = std::upper_bound(xs.begin(), xs.end(), x);
+  // loc - begin guaranteed to be in [0,len]
+  size_t index = (loc - xs.cbegin());
+
+  // Move index and x to be interior
+  size_t len = xs.size();
```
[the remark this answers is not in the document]
attom: Updated

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+  double xmin = xs[0];
+  double xmax = xs[len - 1U];
+  index = std::min(len - 1U, std::max(static_cast<size_t>(1), index)) - 1U;
```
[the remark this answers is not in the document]
attom: Yes, it was unavoidable. Otherwise, get `no matching function for call to 'max(size_t&, unsigned int)'`

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+ * \param[in] xL gridpoint to the left of x (xL <= x)
+ * \param[in] xR gridpoint to the right of x (x <= xR)
+ * \return length-4 array of Hermite polynomials used for interpolation from data at xL and xR to x
+ *
+ * If vL/vR are values of a function at xL/xR and dL/dR are derivatives of a function at xL/xR, then
+ * H[0] * vL + H[1] * vR + H[2] * dL + H[3] * dR
+ * interpolates the value of the function at x, where this function computes and returns H
+ */
+std::array<double, 4> hermite(double x, double xL, double xR) {
+  Require(xL <= x && x <= xR && xL < xR);
```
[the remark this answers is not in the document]
attom: Makes sense. Fixed.

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+ * \param[in] x value of independent variable
+ * \param[in] xL gridpoint to the left of x (xL <= x)
+ * \param[in] xR gridpoint to the right of x (x <= xR)
+ * \return length-4 array of Hermite polynomials used for interpolation from data at xL and xR to x
+ *
+ * If vL/vR are values of a function at xL/xR and dL/dR are derivatives of a function at xL/xR, then
+ * H[0] * vL + H[1] * vR + H[2] * dL + H[3] * dR
+ * interpolates the value of the function at x, where this function computes and returns H
+ */
+std::array<double, 4> hermite(double x, double xL, double xR) {
```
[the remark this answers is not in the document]
attom: I'm not going to move it until we know where it's going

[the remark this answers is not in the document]
attom: I templated it on floating-point type

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+  Require(xL <= x && x <= xR && xL < xR);
+
+  // Spacing of interval
+  double dx = xR - xL;
+  // Left/right basis functions
+  double bL = (xR - x) / dx;
+  double bR = (x - xL) / dx;
```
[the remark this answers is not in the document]
attom: Fixed

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+ *
+ * \param[in] filename Name of the binary csk data file to be read
+ *
+ * This function has rank 0 read the binary csk data file, which fills in the class' data members.
+ * The data members are then broadcast to other MPI ranks to finish their construction.
+ */
+Compton_Native::Compton_Native(const std::string &filename)
+    : num_temperatures_(0U), num_groups_(0U), num_leg_moments_(0U), num_evals_(0U), num_points_(0U),
+      Ts_(0U), Egs_(0U), first_groups_(0U), indexes_(0U), data_(0U), derivs_(0U) {
+  Require(filename.length() > 0U);
```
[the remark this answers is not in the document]
attom: yes, we tend to use trailing underscores for private member data functions

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+  Insist(errcode == 0, "Non-zero errorcode. Exiting.");
+
+  int rank = rtt_c4::node();
+  constexpr int bcast_rank = 0;
+
+  // Broadcast sizes
+  size_t data_size = data_.size();
+  std::array<size_t, 6> pack = {num_temperatures_, num_groups_, num_leg_moments_,
+                                num_evals_,        num_points_, data_size};
+  rtt_c4::broadcast(&pack[0], pack.size(), bcast_rank);
```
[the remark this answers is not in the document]
attom: pack is an `array` and guaranteed to be size 6, right?

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+  num_evals_ = pack[p++];
+  num_points_ = pack[p++];
+  data_size = pack[p++];
+
+  // Derived sizes
+  const auto tsz = static_cast<int>(num_temperatures_);
+  const auto egsz = static_cast<int>(num_groups_ + 1U);
+  const auto fgsz = static_cast<int>(num_temperatures_ * num_groups_);
+  const auto isz = static_cast<int>(fgsz + 1U);
+  const auto dsz = static_cast<int>(data_size);
```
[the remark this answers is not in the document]
attom: I think I needed int's for Draco's MPI broadcast to work? Should never exceed 2 billion data points in practice (if it did, would take 30+ GB for data and derivatives). I'll add a check on data_size to insist it is less than `1U << 31U`

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+
+  // Broadcast grids
+  if (rank != bcast_rank)
+    Ts_.resize(tsz);
+  Check(Ts_.size() == static_cast<size_t>(tsz));
+  rtt_c4::broadcast(&Ts_[0], tsz, 0);
```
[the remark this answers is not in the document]
attom: Will the broadcast correctly fail with nullptr?

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+
+  // Check for valid file stream
+  try {
+    fin.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+  } catch (std::fstream::failure &e) {
+    return 1;
+  }
+
+  // Ensure valid type
+  // (using vector of char to avoid using C-style strings)
```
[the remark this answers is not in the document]
attom: Good idea. Done

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+    fin.read(&actual[i], sizeof(actual[i]));
+  if (!std::equal(expected.begin(), expected.end(), actual.begin())) {
+    std::cerr << "Expecting binary file " << filename << " to start with '";
+    for (char c : expected)
+      std::cerr << c;
+    std::cerr << "' but got '";
+    for (char c : actual)
+      std::cerr << c;
+    std::cerr << "'" << std::endl;
+    return 2;
```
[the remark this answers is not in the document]
attom: Yes, it's so that all of the ranks can die together if rank 0 has a problem with the read

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+    std::cout << "DBG len(first_groups_) " << fgsz << '\n';
+    std::cout << "DBG len(indexes_) " << isz << '\n';
+    std::cout << "DBG len(data_/derivs_) " << dsz << '\n';
+#endif
+
+  Ts_.resize(tsz);
+  for (size_t i = 0; i < tsz; ++i) {
+    // Convert from FP (type in the binary file) to double
+    FP tmp;
+    fin.read(reinterpret_cast<char *>(&tmp), sizeof(FP));
```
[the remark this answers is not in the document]
attom: Right now, both Compton_Native.cc (data reader) and cskrw.cc (data writer) define FP and UINT as the same types, which is required by the file format that is in the software requirements document. Do you want me to make a header file that defines these types and then have both cc files include the header?

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+    derivs_[i] = static_cast<double>(tmp);
+  }
+
+  fin.close();
+
+#if 0
+    std::cout << '\n';
+    std::cout << "-------------------------------------------------\n";
```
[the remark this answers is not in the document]
attom: Yes, perhaps when the code matures and stabilizes more

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+
+//------------------------------------------------------------------------------------------------//
+// TEMPERATURE INTERPOLATION FUNCTIONS
+//------------------------------------------------------------------------------------------------//
+
+//------------------------------------------------------------------------------------------------//
+/*!
+ * \brief Interpolate csk data in temperature and return dense linear inscattering matrix
+ *
+ * \param[inout] inscat The flattened (1D), dense inscattering matrix as a vector
```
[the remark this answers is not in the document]
attom: Fixed

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+ *               Order of inscat is (slow) [moment, group-to, group-from] (fast)
+ *               Does NOT need to be the right size prior to calling
+ * \param[in] Te_keV The electron temperature in keV at which the interpolation is desired
+ * \param[in] num_moments_truncate The maximum number of Legendre moments to use
+ *            Function will use the minimum of this variable and number of moments in the data
+ */
+void Compton_Native::interp_dense_inscat(vec_d &inscat, double Te_keV,
+                                         size_t num_moments_truncate) const {
+  // Finds index and nudges Teff such that
```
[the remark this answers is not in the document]
attom: For robustness, I don't require Te_keV > 0. The `find_index` function deals with extrapolation concerns gracefully. What specifically would you like to see checked by DbC?

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+  // Fill Hermite function
+  std::array<double, 4> hermite = rtt_compton_tools::hermite(Teff, Ts_[iT], Ts_[iT + 1U]);
+
+  // Precompute some sparse indexes
+  const size_t sz = indexes_[indexes_.size() - 1];
+  const size_t end_leg = std::min(static_cast<size_t>(num_moments_truncate), num_leg_moments_);
+  const size_t eval_offset = 0; // in_lin
+
+  // Resize and fill with zeros
+  inscat.resize(end_leg * num_groups_ * num_groups_);
```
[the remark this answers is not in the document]
attom: Wouldn't be the same. `resize` won't overwrite preexisting values and I want to be sure inscat is filled with zeros everywhere (inscat is passed in as a function argument and so could have who knows what)

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+  std::fill(inscat.begin(), inscat.end(), 0.0);
+
+  // Apply Hermite function
+  for (size_t k = 0; k < end_leg; ++k) {
```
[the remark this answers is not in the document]
attom: Yes, I studied the loop ordering in my stand-alone test code and found this was the fastest on x86 procs

src/compton_tools/Compton_Native.hh
```diff
@@ -0,0 +1,220 @@
+//--------------------------------------------*-C++-*---------------------------------------------//
+/*!
+ * \file   compton_tools/Compton_Native.hh
+ * \author Andrew Till
+ * \date   11 May 2020
+ * \brief  Implementation file for native compton bindary-read and temperature interpolation
+ * \note   Copyright (C) 2017-2020 Triad National Security, LLC. All rights reserved.
```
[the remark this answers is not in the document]
attom: Fixed in several places

src/compton_tools/Compton_Native.hh
```diff
@@ -0,0 +1,220 @@
...
+ * \class Compton_Native
+ *
+ * \brief Provides access to relativistic Compton scattering angle and
+ *        multigroup frequency distributions from csk data files
+ *
+ * This interface class allows the client to:
+ * 1) access (interpolate) data from existing multigroup csk libraries
+ * 2) obtain auxiliary information from existing multigroup libraries
+ *    (electron temperature bounds, frequency group structures, etc)
+ *
```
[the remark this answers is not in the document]
attom: fixed

src/compton_tools/Compton_Native.hh
```diff
@@ -0,0 +1,220 @@
...
+//------------------------------------------------------------------------------------------------//
+class Compton_Native {
+
+public:
+  //--------------------------------------------------------------------------//
+  // SETUP FUNCTIONS
+  //--------------------------------------------------------------------------//
+
+  // \brief Constructor
+  Compton_Native(const std::string &filename);
```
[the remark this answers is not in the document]
attom: Added `explicit`. Probably disable?

src/compton_tools/Sparse_Compton_Matrix.cc
```diff
@@ -0,0 +1,50 @@
...
+ * \date   11 May 2020
+ * \brief  Implementation file for compton interface
+ * \note   Copyright (C) 2017-2020 Triad National Security, LLC. All rights reserved.
+ */
+//------------------------------------------------------------------------------------------------//
+
+#include "compton_tools/Sparse_Compton_Matrix.hh"
+#include "c4/global.hh"
+#include "ds++/Assert.hh"
+
```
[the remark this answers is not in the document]
attom: Done

src/compton_tools/Sparse_Compton_Matrix.cc
```diff
@@ -0,0 +1,50 @@
...
+#include <iostream>
+
+namespace rtt_compton_tools {
+
+Sparse_Compton_Matrix::Sparse_Compton_Matrix(size_t num_points, size_t num_groups,
```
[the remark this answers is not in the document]
attom: This is a stub class at present that's commented out of CMakeLists.txt but which may see future action. I'll add a stub autodoc

src/compton_tools/Sparse_Compton_Matrix.hh
attom: Yeah, I'll add `#ifdef 0`

src/compton_tools/cskrw.cc
attom: replaced with default constructor

src/compton_tools/cskrw.cc
attom: I'd like to have a code that in no way requires the CSK library. Added history and intent comments.

src/compton_tools/python/common_compton.py
attom: For reference, all of the python in `src/compton_tools/python` are little scripts intended for internal use so we don't have to keep rewriting our own python readers of Compton data

src/compton_tools/Compton_Native.hh
```diff
@@ -0,0 +1,220 @@
...
+                    true;
+    return all_good;
+  }
+
+private:
+  //--------------------------------------------------------------------------//
+  // PRIVATE DATA
+  //--------------------------------------------------------------------------//
+
+  size_t num_temperatures_;
```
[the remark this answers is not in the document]
attom: Added comments and changed to use default initialization

src/compton_tools/Compton_Native.cc
```diff
@@ -0,0 +1,632 @@
...
+  // Ts_[index] <= Teff <= Ts_[index+1] and 0 <= index <= Ts_.size()-2;
+  double Teff = Te_keV;
+  size_t iT = rtt_compton_tools::find_index(Ts_, Teff);
+
+  // Fill Hermite function
+  std::array<double, 4> hermite = rtt_compton_tools::hermite(Teff, Ts_[iT], Ts_[iT + 1U]);
+
+  // Precompute some sparse indexes
+  const size_t sz = indexes_[indexes_.size() - 1];
+  const size_t end_leg = std::min(static_cast<size_t>(num_moments_truncate), num_leg_moments_);
```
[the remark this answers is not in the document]
attom: It is. Removing

attom:
Alright. Ready for another round of reviewing

keadyk:
> Alright. Ready for another round of reviewing

*rings the bell* FIGHT! :)

I will take one more quick look, but I reckon this is about ready to be smooshed up and merged!

src/compton_tools/Compton_Native.cc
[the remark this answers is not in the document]
keadyk: I agree with @attom, it's a std array, so should be sized at compile time to 6 - always. I don't think the reference to element 0 should be an issue in this context :)

src/compton_tools/Compton_Native.cc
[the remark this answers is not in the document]
keadyk: It *should* just be a pass-through in that case, because it'll get passed a size of zero via tsz :)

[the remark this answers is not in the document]
keadyk: (But I believe @KineticTheory 's original comment about the "&" syntax still stands)

keadyk:
Alright, I went ahead and marked things that seemed cut-and-dried as "resolved" :)

@KineticTheory @attom On the interpolation functions front - I will gladly self-assign the task of moving Andrew's CSK-related (and, eventually, @brryan 's NDI-related) interpolation functions into a shared location in draco (@KineticTheory would adding an "interpolation" directory be okay?).

src/compton_tools/python/csk_reader.py
```diff
@@ -0,0 +1,138 @@
...
+    for ending in endings:
+        if filebase.endswith(ending):
+            fileroot = filebase[:filebase.find(ending)]
+
+    if verbosity:
+        print('Root is', fileroot)
+
+    # normalizations
+    mec2 = 510.998  # keV
+    csk_norm = 2.0 * 0.037558  # currently approximate
```
alexrlongne, line 42: @keadyk Mentioned that this is explained in CSK, do you want to include any of those comments here? Not a big deal.

src/compton_tools/python/common_compton.py
alexrlongne: Looks good, just the small comment about the normalization constant.

alexrlongne (approved):

src/compton_tools/python/csk_reader.py
attom: Good call. Added comments and more digits

KineticTheory:
Replaced by #968

PR #933, opened 2020-10-20
Apply clang-tidy fixes to rng component.

KineticTheory:
### Background

* I am trying to get most of draco up to the requirements of a basic _clang-tidy_ check (nightly regression).
* Update files in _rng_ to they pass the check and activate this directory in the nightly regression system.

### Description of changes

+ Changes:
  - `typedef -> using`
  - use of `=delete`
  - use `auto` when type is specifed on RHS.
  - use `std::array` instead of C-style array.
  - use `emplace_back` instead of `push_back`
  - use `override` decoration
  - use `nullptr` instead of `NULL`
+ Avoid processing with clang-tidy for some tests that are taken directly from R123.
+ Modernize syntax in `tstRnd_Control_Inline.cc`.
+ Also update to 100-column format for comments
+ Also update copyright block for some files.

### Status

* Reference:
* Checks


KineticTheory:
For some reason doxygen was very unhappy after the initial set of changes. I had to update files in _parser_, _cdi_, and _special_functions_ to build the autodoc target without warnings.

KineticTheory:
@alexrlongne If you have time, can you give this a once over review and merge?  I mixed 100-column, clang-format, and clang-tidy changes (I will separate these as much as possible in future PRs).  The changes are straight forward.  Let me know if there are questions.

src/parser/Class_Parser.hh
```diff
@@ -405,15 +360,13 @@ private:
...
-DLL_PUBLIC_parser bool
-Is_Well_Formed_Keyword(Class_Parser_Keyword<Class> const &key);
+template <typename Class> bool Is_Well_Formed_Keyword(Class_Parser_Keyword<Class> const &key);
 
 } // namespace rtt_parser
 
 #include "Class_Parser.i.hh"
 
-#endif // rtt_Parse_Table_HH
+#endif // rtt_Class_Parser__HH
```
alexrlongne, line 369: Very minor: remove extra underscore

src/rng/Counter_RNG.hh
```diff
@@ -296,13 +288,8 @@ public:
...
 private:
-  mutable ctr_type::value_type data[CBRNG_DATA_SIZE];
-
-  //! Private copy constructor.
-  Counter_RNG(const Counter_RNG &);
-
-  //! Private assignment operator.
-  Counter_RNG &operator=(const Counter_RNG &);
+  //! \bug Can this be changed to std::array<ctr_type::value_type,CBRNG_DATA_SIZE> ?
+  mutable ctr_type::value_type data[CBRNG_DATA_SIZE]; // NOLINT
```
alexrlongne, line 292: Just a comment here--I think this was for pre-C++11 support in Random123?

src/rng/test/kat_cpp.cpp
```diff
@@ -98,7 +98,7 @@ void dev_execute_tests(kat_instance *tests, unsigned ntests) {
   unsigned i;
   for (i = 0; i < ntests; ++i) {
     kat_instance *ti = &tests[i];
-    genmap_t::iterator p = genmap.find(make_pair(ti->method, ti->nrounds));
+    auto p = genmap.find(make_pair(ti->method, ti->nrounds));
```
alexrlongne, line 101: I still don't like how clang tidy chooses to go to `auto` here. I thought it was only supposed to do this when they type was obvious from the RHS?

alexrlongne (approved):

alexrlongne:
@KineticTheory My comment are minor, I am curious what you think about how clang-tidy uses `auto` is this something we could set, like in clang-format?

alexrlongne:
Whoops, sorry! Didn't mean to close and report that!

KineticTheory:
> @KineticTheory My comment are minor, I am curious what you think about how clang-tidy uses `auto` is this something we could set, like in clang-format?

Thanks for reviewing these changes.  I'll work on adjustments this morning.  

I'll take another look at the use of `auto`.  I can't remember if that was a `clang-tidy` recommendation or some other issue.

It looks like I need a rebase as well.

src/rng/Counter_RNG.hh
KineticTheory: > Just a comment here--I think this was for pre-C++11 support in Random123?

I think you are correct.  But I didn't want to figure out how to convert the type to `std::array` right now.  We can revisit this in the future.

src/rng/test/kat_cpp.cpp
KineticTheory: After double checking, I confirmed that `clang-tidy` doesn't like this.  Here's the rule:
```
.../src/rng/test/kat_cpp.cpp:101:5: warning: use auto when declaring iterators [modernize-use-auto]
    genmap_t::iterator p = genmap.find(make_pair(ti->method, ti->nrounds));
    ^~~~~~~~~~~~~~~~~~
    auto
```

I'm sympathetic to your concern, but in this case I think `auto` actually makes sense. The actual data type is defined 5 lines above and the scope of use in limited to the next 3 lines of code.  That actual data type isn't really descriptive in this context.  I think I will leave it as `auto`.

src/parser/Class_Parser.hh
KineticTheory: Fixed in 6349602.

alexrlongne:
I'm good to merge, I really like most of the `modernizes-use-auto` flags (full documentation here: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html ).  I was just a little confused because it does more than just use `auto` when the type is specified in the RHS.

PR #957, opened 2020-11-09
Nvcc wrapper support

david8dixon:
### Background

* nvcc_wrapper preprocesses everything with nvcc. As such, any cpu code that includes the previous version of Assert.hh and uses Insist was getting the version intended for device code. This causes API breaks when compiling with the wrapper were strings are generally passed to this function in most cpu code.

### Purpose of Pull Request

* Add another macro that is explicitly for use in device code until kokkos developers fix the wrapper.

### Description of changes

* One will find general replacement of __NVCC__ with __CUDA_ARCH__ which was thought to correct the issue before we fully understood what the root cause was. However this change remained as it is more general.
* Added the Insist_device macro which is an alias to no_exception_insist and is only for use in device code.

### Status
This was tested by hand when in a build context and we also tested a deployment in client code.

* Reference:
* Checks

### DIFF
--- a/src/ds++/Assert.hh
+++ b/src/ds++/Assert.hh
@@ -333,11 +333,10 @@ std::string verbose_error(std::string const &message);
 // Always on
 //------------------------------------------------------------------------------------------------//
 #if ( defined __CUDA_ARCH__  && defined USE_CUDA )
-#define Insist(c, m) if(!(c)) rtt_dsxx::no_exception_insist( #c, m, __FILE__, __LINE__)
-#else
+#define Insist_device(c, m) if(!(c)) rtt_dsxx::no_exception_insist( #c, m, __FILE__, __LINE__)
+#endif
 #define Insist(c,m) if (!(c)) rtt_dsxx::insist( #c, m, __FILE__, __LINE__ )
 #define Insist_ptr(c,m) if (!(c)) rtt_dsxx::insist_ptr( #c, m, __FILE__, __LINE__ )
-#endif
 
 #elif DBC & 16
 
@@ -490,6 +489,9 @@ std::string verbose_error(std::string const &message);
 #endif
 #ifndef Insist
 #define Insist(c,m)
+#endif
+#ifndef Insist
+#define Insist_device(c,m)
 #endif
 #ifndef Insist_ptr
 #define Insist_ptr(c,m)
--- a/src/ds++/Assert.hh
+++ b/src/ds++/Assert.hh
@@ -176,7 +176,7 @@ void show_cookies(std::string const &cond, std::string const &file, int const li
 [[noreturn]] void insist_ptr(char const *const cond, char const *const msg, char const *const file,
                              int const line);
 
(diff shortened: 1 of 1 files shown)


KineticTheory:
I will defer to @alexrlongne on these changes.  I thought he already had a DbC implementation (possibly located in `device`) that works correctly on the GPU.  If so, and we adopt this solution, then we will need remove portions of `device` and update Jayenne to work with the new model.  This may take some time.

src/ds++/Assert.hh
```diff
@@ -332,12 +332,11 @@ std::string verbose_error(std::string const &message);
 //------------------------------------------------------------------------------------------------//
 // Always on
 //------------------------------------------------------------------------------------------------//
-#if ( defined __NVCC__  && defined USE_CUDA )
-#define Insist(c, m) if(!(c)) rtt_dsxx::no_exception_insist( #c, m, __FILE__, __LINE__)
-#else
+#if ( defined __CUDA_ARCH__  && defined USE_CUDA )
+#define Insist_device(c, m) if(!(c)) rtt_dsxx::no_exception_insist( #c, m, __FILE__, __LINE__)
+#endif
```
alexrlongne, line 337: What do you think about adding an `else` here that defines `Insist_device` to be the CPU version? That way if I change the shared CPU/GPU code in Jayenne to use `Insist_device` they will still get the correct throwing version in CPU code. This should also solve the draco problem because all of draco uses the standard `Insist` and there won't be any confusion in which function is called under the hood.

alexrlongne:
> I will defer to @alexrlongne on these changes. I thought he already had a DbC implementation (possibly located in `device`) that works correctly on the GPU. If so, and we adopt this solution, then we will need remove portions of `device` and update Jayenne to work with the new model. This may take some time.

Draco's `Assert` has the no-throw GPU version but because of the KOKKOS preprocessor step, macros are defined that lead to the GPU version being included in CPU code (the CPU version allows `std::string`, which is not allowed in device code and fails to compile). I can make the changes in Jayenne pretty easily because a "no_exception_insist" should only be used in code that could be called on the device (in `tracking` or `imc_solver`).

src/ds++/Assert.hh
david8dixon: Good catch. I will add that back in. That should work because I c strings will convert to std::strings just not the other way around.

david8dixon:
I addressed @alexrlongne comment but we do not have any tests in Draco to verify any of this works. Not sure if that is something we want to add as a part of this PR. Let me know.

alexrlongne:
> I addressed @alexrlongne comment but we do not have any tests in Draco to verify any of this works. Not sure if that is something we want to add as a part of this PR. Let me know.

I'm fine with it as is since everything should default to the standard, tested `Insist` for all the cases in draco. I need to clean up the `device` directory at some point and I'll add some tests for the draco GPU code I actually use at that point.

KineticTheory (approved):
Thanks for working with Alex and Jeff to improve this code.

PR #979, opened 2020-12-10
Make global_sum more generic

KineticTheory:
### Background

* As we increase the compile warning levels in our code we are starting to see issues with `global_sum(T* data, int n)` because we are passing integral types for `n` that are different than `int` (e.g. `uint64_t`).  Instead of creating a bunch of overloaded signatures, use a template to represent the size.  But only all the type of `n` to be an integral type.
* If this works well, we might consider updating other parts of `c4` to use a similar signature.

### Description of changes

* Add a dummy 3rd argument to the function signature that provides a mechanism for limiting the implicit instantiation to integral types for the array length
* Didn't mean to include those changes to `.bashrc_cray`, but they are extremely minor and I'm too lazy to reset the PR.

### Status

* Reference:
* Checks


src/c4/C4_Functions.hh
```diff
@@ -352,8 +352,8 @@ template <typename T> void global_max(T &x);
 /*!
  * \brief Do an element-wise, global sum of an array.
  */
-template <typename T> void global_sum(T *x, int n);
-template <typename T> void global_sum(T *x, size_t n);
+template <typename T, typename L>
+void global_sum(T *x, L n, typename std::enable_if<std::is_integral<L>::value, L>::type * = 0);
```
alexrlongne, line 356: Oh man, this is hard to parse. I'm ok to merge this but we should probably have some comment around it to make it clear why the third argument is necessary.

alexrlongne (approved):

src/c4/C4_Functions.hh
KineticTheory: @along Good point.  Sorry - its a bit of a rush job.  Additionally, we should probably transform the remaining reductions in the same way.

alexrlongne: No problem! I like that it made the instantiations go away. I'm interested to learn more about `enable_if`.

KineticTheory: Do you prefer this form?

```c++
/*!
 * \brief Do an element-wise, global sum of an array.
 * 
 * Only instantiate this function for types L that integral.  Details of this pattern are at
 * https://en.cppreference.com/w/cpp/types/enable_if.
 */
template <typename T, typename L,
          typename std::enable_if<std::is_integral<L>::value, bool>::type = true>
void global_sum(T *x, L n);
````

PR #981, opened 2020-12-11
Provide an alternate implementation for global_sum

KineticTheory:
### Background

+ This approach is a little easier to parse the the previous solution merged in #979.
+ Also fixes 3 clang-tidy issues.

### Status

* Reference:
* Checks

### DIFF
--- a/src/c4/C4_Functions.hh
+++ b/src/c4/C4_Functions.hh
@@ -351,9 +351,13 @@ template <typename T> void global_max(T &x);
 //------------------------------------------------------------------------------------------------//
 /*!
  * \brief Do an element-wise, global sum of an array.
+ * 
+ * Only instantiate this function for types L that are integral. Details of this pattern are at
+ * https://en.cppreference.com/w/cpp/types/enable_if.
  */
-template <typename T, typename L>
-void global_sum(T *x, L n, typename std::enable_if<std::is_integral<L>::value, L>::type * = 0);
+template <typename T, typename L,
+          typename std::enable_if<std::is_integral<L>::value, bool>::type = true>
+void global_sum(T *x, L n);
 
 //------------------------------------------------------------------------------------------------//
 /*!
--- a/src/c4/C4_MPI.i.hh
+++ b/src/c4/C4_MPI.i.hh
@@ -17,9 +17,8 @@
 namespace rtt_c4 {
 
 //------------------------------------------------------------------------------------------------//
-
-template <typename T, typename L>
-void global_sum(T *x, L n, typename std::enable_if<std::is_integral<L>::value, L>::type *) {
+template <typename T, typename L, typename std::enable_if<std::is_integral<L>::value, bool>::type>
+void global_sum(T *x, L n) {
   Require(x != nullptr);
   Require(n > 0);
   Require(n < INT32_MAX);
--- a/src/c4/C4_Serial.hh
+++ b/src/c4/C4_Serial.hh
@@ -42,9 +42,8 @@ int create_vector_type(unsigned /*count*/, unsigned /*blocklength*/, unsigned /*
 // Global_<Op> functions
 //------------------------------------------------------------------------------------------------//
 
-template <typename T, typename L>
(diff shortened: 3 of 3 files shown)


KineticTheory:
@alexrlongne I accidentally deleted the branch associated with #980.  This replaces the other PR.

src/c4/C4_Functions.hh
```diff
@@ -351,9 +351,13 @@ template <typename T> void global_max(T &x);
 //------------------------------------------------------------------------------------------------//
 /*!
  * \brief Do an element-wise, global sum of an array.
+ * 
+ * Only instantiate this function for types L that are integral. Details of this pattern are at
+ * https://en.cppreference.com/w/cpp/types/enable_if.
  */
-template <typename T, typename L>
-void global_sum(T *x, L n, typename std::enable_if<std::is_integral<L>::value, L>::type * = 0);
```
alexrlongne, line 356: Do you need more than the is <is_integral<L>::value? Example 5 on the enable_if page looks like it should work?
```
// #5, enabled via a type template parameter
template<class T,
	typename = std::enable_if_t<std::is_array<T>::value> >
void destroy(T* t) // note: function signature is unmodified
{
    for(std::size_t i = 0; i < std::extent<T>::value; ++i) {
        destroy((*t)[i]);
    }
}
```

KineticTheory: The example in #5 is more complicated to understand but the syntax is simpler.  Initial testing shows that the following model will work:

```c++
// C4_Functions.hh
template <typename T, typename L, typename = std::enable_if<std::is_integral<L>::value>>
void global_sum(T * x, L n);
```

```c++
// C4_MPI.i.hh
template <typename T, typename L, typename Enabled>
void global_sum(T *x, L n) {
  Require(x != nullptr);
  Require(n > 0);
  Require(n < INT32_MAX);
  MPI_Allreduce(MPI_IN_PLACE, x, static_cast<int>(n), MPI_Traits<T>::element_type(), MPI_SUM, communicator);
}
```

What is your opinion on this solution?

alexrlongne: I think I prefer your original solution as it has everything in one place. Sorry about the red herring with that other example.

KineticTheory: Actually, this alternate version appears to be broken since it instantiates a version for `L=double`.

KineticTheory: > 
> 
> I think I prefer your original solution as it has everything in one place. Sorry about the red herring with that other example.

All of these implementations have separate declaration and implementation because of the `DRACO_C4=MPI|SCALAR` option (even the fix that we merged yesterday).  I think I like the following a bit better than the existing implementation because it eliminates the extra 3rd argument by using a 3rd template argument.

```c++
// C4_Functions.hh
/*!
 * \brief Do an element-wise, global sum of an array.
 * 
 * Only instantiate this function for types L that are integral. Details of this pattern are at
 * https://en.cppreference.com/w/cpp/types/enable_if.
 */
template <typename T, typename L,
          typename std::enable_if<std::is_integral<L>::value, bool>::type = true>
void global_sum(T *x, L n);
```

```c++
// C4_MPI.i.hh (similar signature in C4_Serial.hh)
template <typename T, typename L, typename std::enable_if<std::is_integral<L>::value, bool>::type>
void global_sum(T *x, L n) {
  Require(x != nullptr);
  Require(n > 0);
  Require(n < INT32_MAX);
  MPI_Allreduce(MPI_IN_PLACE, x, static_cast<int>(n), MPI_Traits<T>::element_type(), MPI_SUM,
                communicator);
}
```

Comments?

KineticTheory: (I just realized that I re-derived my first solution from this morning.)  
Do you prefer to have the type-toggle for implementation in (1) the template list, or (2) the argument list?

alexrlongne: Sorry, I probably said the wrong thing with "original". I like the template list (1) personally!

PR #984, opened 2020-12-14
Save ndi version to cdi_ndi/config.h

KineticTheory:
### Background

* Update build system to scrape NDI version and save it to `cdi_ndi/config.h`.
* Produces something like this:
```c++
#define NDI_VERSION_STRING "2.2.0alpha"
#define NDI_MAJOR "2"
#define NDI_MINOR "2"
```
* Note that NDI itself defines:
```
#define NDI_VERSION_DATA "2.2.0alpha"
...
# define NDI_VERSION 42
```
* We must be careful to avoid symbol collision.

### Status

* Reference:
* Checks
* WIP
  * [x] Fix warning ` warning #47: incompatible redefinition of macro "NDI_VERSION"`

### DIFF
--- a/config/FindNDI.cmake
+++ b/config/FindNDI.cmake
@@ -62,8 +62,7 @@ include(FindPackageHandleStandardArgs)
 # at this location over system locations.
 if( EXISTS "$ENV{NDI_ROOT_DIR}" )
   file( TO_CMAKE_PATH "$ENV{NDI_ROOT_DIR}" NDI_ROOT_DIR )
-  set( NDI_ROOT_DIR "${NDI_ROOT_DIR}" CACHE PATH
-    "Prefix for NDI installation." )
+  set( NDI_ROOT_DIR "${NDI_ROOT_DIR}" CACHE PATH "Prefix for NDI installation." )
 endif()
 
 #=============================================================================
@@ -77,23 +76,33 @@ endif()
 find_path( NDI_INCLUDE_DIR
   NAMES ndi.h
   HINTS ${NDI_ROOT_DIR}/include/ndi ${NDI_ROOT_DIR}/include
-  PATH_SUFFIXES ndi
-  )
-
+  PATH_SUFFIXES ndi )
 set( NDI_LIBRARY_NAME ndipic ndi)
 
 find_library(NDI_LIBRARY
   NAMES ${NDI_LIBRARY_NAME}
-  PATHS ${NDI_ROOT_DIR}/lib
-  )
+  PATHS ${NDI_ROOT_DIR}/lib )
 # Do we also have debug versions?
 find_library( NDI_LIBRARY_DEBUG
   NAMES ${NDI_LIBRARY_NAME}
   HINTS ${NDI_ROOT_DIR}/lib
-  PATH_SUFFIXES Debug
-)
+  PATH_SUFFIXES Debug )
 set( NDI_INCLUDE_DIRS ${NDI_INCLUDE_DIR} )
 set( NDI_LIBRARIES ${NDI_LIBRARY} )
+
+# Try to find the version.
+if( NOT NDI_VERSION )
+  # We might also try scraping the directory name for a regex match "ndi-X.X.X"
+  if( NOT NDI_MAJOR )
+    string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\1" NDI_MAJOR
+      ${NDI_INCLUDE_DIR} )
+    string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\2" NDI_MINOR
+      ${NDI_INCLUDE_DIR} )
+    string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\3" NDI_SUBMINOR
+      ${NDI_INCLUDE_DIR} )
+    set( NDI_VERSION "${NDI_MAJOR}.${NDI_MINOR}.${NDI_SUBMINOR}")
+  endif()
+endif()
 
 #=============================================================================
 # handle the QUIETLY and REQUIRED arguments and set NDI_FOUND to TRUE if
@@ -105,11 +114,10 @@ find_package_handle_standard_args( NDI
     NDI_INCLUDE_DIR
     NDI_LIBRARY
   VERSION_VAR
-    NDI_VERSION
-    )
+    NDI_VERSION )
 
-mark_as_advanced( NDI_ROOT_DIR NDI_VERSION NDI_LIBRARY
-  NDI_INCLUDE_DIR NDI_LIBRARY_DEBUG NDI_USE_PKGCONFIG NDI_CONFIG )
+mark_as_advanced( NDI_ROOT_DIR NDI_VERSION NDI_LIBRARY NDI_INCLUDE_DIR NDI_LIBRARY_DEBUG
+    NDI_USE_PKGCONFIG NDI_CONFIG )
 
 #=============================================================================
 # Register imported libraries:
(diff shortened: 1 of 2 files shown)


brryan:
@KineticTheory Thanks for adding this -- for non-standard installs (like my local install) I get `NDI_VERSION`s like `/home/brryan/xcp-stash/ndi/install/include./home/brryan/xcp-stash/ndi/install/include./home/brryan/xcp-stash/ndi/install/include` for `NDI_ROOT_DIR=/home/brryan/xcp-stash/ndi/install/`. What about this check for the directory format we plan on parsing, which reports the version as `Unknown` if `NDI_INCLUDE_DIR` doesn't follow contain the version info in the expected format?
```cmake
# Try to find the version.                                                                           
if( NOT NDI_VERSION )                                                                                
  # We might also try scraping the directory name for a regex match "ndi-X.X.X"                      
  if( NOT NDI_MAJOR )                                                                                
    message(STATUS ${NDI_INCLUDE_DIR})                                                               
    if (NDI_INCLUDE_DIR MATCHES "ndi-([0-9]+).([0-9]+).([0-9]+)")                                    
      string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\1" NDI_MAJOR               
        ${NDI_INCLUDE_DIR} )                                                                         
      string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\2" NDI_MINOR               
        ${NDI_INCLUDE_DIR} )                                                                         
      string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\3" NDI_SUBMINOR            
        ${NDI_INCLUDE_DIR} )                                                                         
      set( NDI_VERSION "${NDI_MAJOR}.${NDI_MINOR}.${NDI_SUBMINOR}")                                  
    else()                                                                                           
      set( NDI_MAJOR "Unknown" )                                                                     
      set( NDI_MINOR "Unknown" )                                                                     
      set( NDI_SUBMINOR "Unknown" )                                                                  
      set( NDI_VERSION "Unknown" )                                                                   
    endif()                                                                                          
  endif()                                                                                            
endif()
```

KineticTheory:
> 
> 
> @KineticTheory Thanks for adding this -- for non-standard installs (like my local install) I get `NDI_VERSION`s like `/home/brryan/xcp-stash/ndi/install/include./home/brryan/xcp-stash/ndi/install/include./home/brryan/xcp-stash/ndi/install/include` for `NDI_ROOT_DIR=/home/brryan/xcp-stash/ndi/install/`. What about this check for the directory format we plan on parsing, which reports the version as `Unknown` if `NDI_INCLUDE_DIR` doesn't follow contain the version info in the expected format?
> 
> ```cmake
> # Try to find the version.                                                                           
> if( NOT NDI_VERSION )                                                                                
>   # We might also try scraping the directory name for a regex match "ndi-X.X.X"                      
>   if( NOT NDI_MAJOR )                                                                                
>     message(STATUS ${NDI_INCLUDE_DIR})                                                               
>     if (NDI_INCLUDE_DIR MATCHES "ndi-([0-9]+).([0-9]+).([0-9]+)")                                    
>       string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\1" NDI_MAJOR               
>         ${NDI_INCLUDE_DIR} )                                                                         
>       string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\2" NDI_MINOR               
>         ${NDI_INCLUDE_DIR} )                                                                         
>       string( REGEX REPLACE ".*ndi-([0-9]+).([0-9]+).([A-Za-z0-9]+).*" "\\3" NDI_SUBMINOR            
>         ${NDI_INCLUDE_DIR} )                                                                         
>       set( NDI_VERSION "${NDI_MAJOR}.${NDI_MINOR}.${NDI_SUBMINOR}")                                  
>     else()                                                                                           
>       set( NDI_MAJOR "Unknown" )                                                                     
>       set( NDI_MINOR "Unknown" )                                                                     
>       set( NDI_SUBMINOR "Unknown" )                                                                  
>       set( NDI_VERSION "Unknown" )                                                                   
>     endif()                                                                                          
>   endif()                                                                                            
> endif()
> ```

I would expect your use case to be caught in code by `#ifndef NDI_VERSION`.  

HOWEVER - I just realized `NDI_INCLUDE_DIR/include/ndi_version.h` provides `#define NDI_VERSION_DATA`.  Let me update the cmake script to fetch this value instead of scraping the information from the path.

KineticTheory:
@brryan I think 06a22dd will work for your use case.  My only concern is that I'm not sure if all older versions of NDI provide this CPP macro.  If the file is missing, I will still attempt to pull the version from the filepath.  If that fails, then the NDI_VERSION CPP macro will not be defined and can use logic like:
```c++
#ifdef NDI_VERSION
#  if NDI_MAJOR > 2 && NDI_MINOR > 2
     v="good"
#  else
     v="too old"
#  endif
#else
   v = "unknown"
#endif
```

keadyk:
Sorry @brryan and @KineticTheory, I think I missed discussion on this at the meeting this morning! Are we going to use the version info to trigger different behavior in the code, disable the NDI capability altogether (ie set NDI_FOUND=no), or some third option? :)

KineticTheory:
> 
> 
> Sorry @brryan and @KineticTheory, I think I missed discussion on this at the meeting this morning! Are we going to use the version info to trigger different behavior in the code, disable the NDI capability altogether (ie set NDI_FOUND=no), or some third option? :)

Sorry - yes, we discussed this at the planning meeting this morning.  The issue is that we have introduced dependencies on ndi-2.2.0+ to support reading stopping power data that prevent us from linking to older versions of ndi.  Ben will try to create some version-dependent linking so that only the stopping power data reader is disabled when linking to older versions.  This will allow us to continue using older ndi versions for the next 6 months or so w/o disabling all of CPT.

brryan:
Thanks @KineticTheory! This looks like a really robust solution, and works as expected on my end. I'm not familiar with the inception data of `ndi_version.h` but I would be surprised if both that file's macro doesn't exist/can't be manually created by a sufficiently sneaky Draco developer AND the install directory doesn't have the format `ndi-*.*.*`.

config/FindNDI.cmake
```diff
@@ -77,24 +76,34 @@ endif()
...
 find_library( NDI_LIBRARY_DEBUG
   NAMES ${NDI_LIBRARY_NAME}
   HINTS ${NDI_ROOT_DIR}/lib
-  PATH_SUFFIXES Debug
-)
+  PATH_SUFFIXES Debug )
 set( NDI_INCLUDE_DIRS ${NDI_INCLUDE_DIR} )
 set( NDI_LIBRARIES ${NDI_LIBRARY} )
 
+# Try to find the version.
```
brryan, line 93: @KineticTheory If you add the 06a22dd change from the discussion I will happily approve :)

KineticTheory: It's already there!

keadyk:
> > Sorry @brryan and @KineticTheory, I think I missed discussion on this at the meeting this morning! Are we going to use the version info to trigger different behavior in the code, disable the NDI capability altogether (ie set NDI_FOUND=no), or some third option? :)
> 
> Sorry - yes, we discussed this at the planning meeting this morning. The issue is that we have introduced dependencies on ndi-2.2.0+ to support reading stopping power data that prevent us from linking to older versions of ndi. Ben will try to create some version-dependent linking so that only the stopping power data reader is disabled when linking to older versions. This will allow us to continue using older ndi versions for the next 6 months or so w/o disabling all of CPT.

Cool! That sounds great. Let me know if there's anything I can do to help :)

config/FindNDI.cmake
brryan: Weird -- last night github's web interface was saying something like "This commit is not a member of any PR". I see it now though!

brryan:
LGTM!

brryan (approved):

KineticTheory:
@brryan There were symbol conflicts on snow.  Let's wait for an updated CI report before merging.

PR #1001, opened 2020-12-31
Use CUDA_ARCHITECTURES CMake property.

KineticTheory:
### Background

+ Starting with CMake-3.18, a warning is issued for every target that is created with `nvcc` and that is missing the property `CUDA_ARCHITECTURES`.

### Purpose of Pull Request

* [Fixes Redmine Issue #2026](https://rtt.lanl.gov/redmine/issues/2026)

### Description of changes

+ Update `Draco_std_target_props_CUDA` defined and used in `component_macros.cmake` so that the required target property is set properly.
+ Eliminate the build variable `Draco_CUDA_ARCH` by using `CUDA_ARCHITECTURES` instead.

### Status

* Reference:
* Checks


KineticTheory:
@alexrlongne I'm going to merge this PR w/o review to eliminate build warnings in the CI. However, I still want you to review this change.  It can be reverted if there is a better way.

config/unix-cuda.cmake
```diff
@@ -18,23 +22,24 @@ include_guard(GLOBAL)
 # - https://devblogs.nvidia.com/building-cuda-applications-cmake/
 
 # Discover CUDA compute capabilities.
-if( NOT DEFINED Draco_CUDA_ARCH )
+if( NOT DEFINED CUDA_ARCHITECTURES )
```
alexrlongne, line 25: Is the `query_gpu` path still here for older versions of CMake?

KineticTheory: As far as I can tell, cmake still doesn't actually do that introspection.  Please let me know if I am wrong about this.

config/unix-cuda.cmake
```diff
@@ -7,6 +7,10 @@
 
 include_guard(GLOBAL)
 
+if(NOT CMAKE_CUDA_COMPILER_ID STREQUAL NVIDIA)
```
alexrlongne, line 10: I think this is redundant since it's checked before the `unix-cuda.cmake` file is included. From `compilerEnv.cmake`:

```
781     # setup flags                                                                
782     if( "${CMAKE_CUDA_COMPILER_ID}" MATCHES "NVIDIA" )                           
783       include( unix-cuda )                                                       
784     else()                                                                       
785       message(FATAL_ERROR "Build system does not support "                       
786         "CUDACXX=${CMAKE_CUDA_COMPILER}")                                        
787     endif()                                                                      
788                          
```

config/unix-cuda.cmake
alexrlongne: OK, got it, I see that after looking at more of the documentation.

It looks like `CMAKE_CUDA_ARCHITECTURES` is supposed to be initialized with a default value but it doesn't seem to be (I tried printing it during configure) so I think it makes sense to leave in the logic to print the value for now.

PR #1008, opened 2021-01-06
Ofpstream

kgbudge:
### Background

* While working on a parallel version of an output option in a code built on Draco, I found that some output routed through an ofpstream was being duplicated. 
* Inspection of the Draco rtt_c4::ofpstream class showed that it was not properly handling the corner case where one or more ranks send() an empty output string.
* The result was that the root rank wrote the output string from the previous rank to the output file a second time.

### Purpose of Pull Request

* Fix rtt_c4::ofpstream for the corner case of some ranks having no output in a send().
* Add a test problem that would have detected this corner case failure.

### Description of changes

* Ensure that if the root rank is sent no output by another rank on a send() call, the receive buffer is cleared. Otherwise the root rank will write whatever is in the buffer (usually output from the previous rank) to the output file a second time.
* Add a test case to tstofpstream.cc that would have detected the failure of this corner case. This incidentally improves testing of the ofpstream class generally.

### Status

* Reference:
* Checks


src/c4/ofpstream.cc
```diff
@@ -52,8 +52,8 @@ void ofpstream::mpibuf::send() {
     for (unsigned i = 1; i < pids; ++i) {
       unsigned N(0);
       receive(&N, 1, i);
+      buffer_.resize(N);
```
kgbudge, line 55: Move the resize call so that it is always made, even when the rank has no output (N==0).

attom:
This LGTM. Not sure how I do an official 'review' on github.

attom (approved):
I think I found the button to press. LGTM

KineticTheory:
I've asked @keadyk and @clevelam to review this PR since they are using this functionality in our code.

KineticTheory (approved):

keadyk (approved):
LGTM - thanks for catching this, Kent! I feel like we may have seen this pop up in our standalone driver (but didn't realize the root cause)

PR #1024, opened 2021-02-03
Extend policy for CDI integrations

kgbudge:
### Background

* There are situations (such as when using opacity tables or a temperature floor) where a client does not have freedom to set his group bounds structure such that it is guaranteed to take in the full Planck or Rosseland integrals. 
* For example, the lowest group bound cannot always be set to zero. This means that part of the integrals "disappears" into the gap below the lowest group bound, particularly at cold temperatures.

### Purpose of Pull Request

* Add a new policy flag to CDI, CDI::extend, that flags the code to add the lower tail to the first group and the upper tail to the last group. This guarantees that the sum of the group values equals 1, to the precision of the integration algorithm.

### Description of changes

* Add static bool CDI::extend to CDI.
* Add a setter, CDI::setExtend() that turns on this flag (which is off by default, preserving legacy behavior)
* Add a getter, CDI::getExtend(), to see if this policy has been set.
* Modify the four functions that integrate Planck or Rosseland functions over a group structure compute the extended values when the extend policy is set.

Note that this does not affect the opacity averaging routines, nor any routines that take individual group bounds rather than a complete group bound vector as their argument. In the former case, the old behavior should always be appropriate. In the latter case, the function cannot know whether it is looking at the lowest or highest group of some group structure.

### Status

* Reference:
* Checks

### DIFF
--- a/src/cdi/CDI.cc
+++ b/src/cdi/CDI.cc
@@ -51,7 +51,7 @@ CDI::~CDI() = default;
 //------------------------------------------------------------------------------------------------//
 
 std::vector<double> CDI::frequencyGroupBoundaries = std::vector<double>();
-bool CDI::extend = false;
+DLL_PUBLIC_cdi bool CDI::extend = false;
 
 //------------------------------------------------------------------------------------------------//
 // STATIC FUNCTIONS
--- a/src/cdi/CDI.cc
+++ b/src/cdi/CDI.cc
@@ -51,6 +51,7 @@ CDI::~CDI() = default;
 //------------------------------------------------------------------------------------------------//
 
 std::vector<double> CDI::frequencyGroupBoundaries = std::vector<double>();
+bool CDI::extend = false;
 
 //------------------------------------------------------------------------------------------------//
 // STATIC FUNCTIONS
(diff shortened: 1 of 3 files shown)


src/cdi/CDI.cc
```diff
@@ -51,6 +51,7 @@ CDI::~CDI() = default;
 //------------------------------------------------------------------------------------------------//
 
 std::vector<double> CDI::frequencyGroupBoundaries = std::vector<double>();
+bool CDI::extend = false;
```
KineticTheory, line 54: Sigh. Global variables always seem to cause problems.  Can you try prepending this line with `DLL_PUBLIC_cdi`? See `c4/C4_MPI.hh` for an example (search for `DLL_PUBLIC_c4`).

src/cdi/CDI.hh
```diff
@@ -511,6 +511,9 @@ class CDI {
   //! Material ID.
   std_string matID;
 
+  //! Extend integration to place low and high tails in low and high groups?
+  static bool extend;
```
KineticTheory: Sigh. Global variables always seem to cause problems.  Can you try prepending this line with `DLL_PUBLIC_cdi`? See `c4/C4_MPI.hh` for an example (search for `DLL_PUBLIC_c4`).

kgbudge: Done.

KineticTheory: If this PR hits additional MSVC issues, I'll checkout your branch and work through the necessary changes.  With any luck it won't come to this.

KineticTheory (approved):
LGTM. Merging.

PR #1036, opened 2021-03-04
Fix CDash URLs in Draco's PR template

alexrlongne:
### Background

* TRT teams are using a new cdash server but the template URLs were not updated.  Fixing it now.


alexrlongne:
@KineticTheory Did I use the right version of the new CDash URLs here?

KineticTheory:
> 
> 
> @KineticTheory Did I use the right version of the new CDash URLs here?

That looks right to me.  Sorry for missing this.

PR #1050, opened 2021-04-11
Update ApplicationUnitTest to support multiple args for the same test.

KineticTheory:
### Background

+ Previously, we only allowed a single command line argument to be assigned during test registration.  This update allows options like '-a -b -c' to be specified for a single test.

### Description of changes

+ Apply new formatting rules for python (flake8) and cmake (cmake-format/cmake-lint).
+ Fix an indentation error in README.md.
+ Fix an issue related to 'draco_info' where the 'build date' was not captured correctly.

### Status

* Reference:
* Checks

### DIFF
--- a/config/ApplicationUnitTest.cmake
+++ b/config/ApplicationUnitTest.cmake
@@ -265,7 +265,7 @@ macro( aut_register_test )
   # Set arguments that don't change between python and CMake driver
   set (SHARED_ARGUMENTS
       -DAPP=${aut_APP}
-      -DARGVALUE=${argvalue}
+      -DARGVALUE=_av_${argvalue}_av_
       -DWORKDIR=${aut_WORKDIR}
       -DTESTNAME=${ctestname_base}${argname}
       -DDRACO_CONFIG_DIR=${DRACO_CONFIG_DIR}
@@ -520,6 +520,7 @@ macro( aut_runTests )
     string( REPLACE ".out" "-${numPE}.out" OUTFILE ${OUTFILE} )
     string( REPLACE ".err" "-${numPE}.err" ERRFILE ${ERRFILE} )
   endif()
+  string(REGEX REPLACE "_av_([A-z-]*)_av_" "\\1" ARGVALUE "${ARGVALUE}" )
   if( ARGVALUE )
     string( REGEX REPLACE "[-]" "" safe_argvalue ${ARGVALUE} )
     string( REPLACE ".out" "-${safe_argvalue}.out" OUTFILE ${OUTFILE} )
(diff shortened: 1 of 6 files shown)


config/application_unit_test.py
```diff
@@ -1,770 +1,768 @@
...
+    re_mpiexec = re.compile(r"MPIEXEC_EXECUTABLE=([^\s]*)")
+    re_mpi_cores_per_cpu = re.compile(r"MPI_CORES_PER_CPU=([^\s]*)")
+    re_draco_info = re.compile(r"DRACO_INFO=([^\s]*)")
+    re_gdiff = re.compile(r"GDIFF=([^\s]*)")
+    re_pgdiff = re.compile(r"PGDIFF=([^\s]*)")
+
+    # run command can include spaces, modfiy regular expression for that
+    re_run_cmd = re.compile("RUN_CMD=(.*?)-D")
+
+    re_arg_value = re.compile("ARGVALUE=_av_(.*)_av_")
```
KineticTheory, line 150: * This line is the only design change in this file.
* However, all of these regular expressions needed the `r` (raw string) qualifier to pass flake8 checks.

config/ApplicationUnitTest.cmake
```diff
@@ -265,7 +265,7 @@ macro( aut_register_test )
   # Set arguments that don't change between python and CMake driver
   set (SHARED_ARGUMENTS
       -DAPP=${aut_APP}
-      -DARGVALUE=${argvalue}
+      -DARGVALUE=_av_${argvalue}_av_
```
KineticTheory, line 268: @alexrlongne Can you think of a better way to do this.  I want to capture the string `-a -b -c` as a single argument.  I couldn't figure out how to make that work for both CMake and Python regex, so I added `_av_` as begin() and end() delimiters.

alexrlongne: For python I can do something like this:
```
 arg = re.compile("(-[A-Za-z0-9]*)"
 my_str = "-option1 -option2"
 arg.findall(my_str)
```

and it returns a nice list:
```
['-option1', '-option2']
```

Let me try a few things in CMake and see if it has something nice like that.

Is there a simple interactive shell for cmake?  :)

alexrlongne: Resolving this for now, we can revisit regexes later

alexrlongne (approved):

config/application_unit_test.py
alexrlongne: OK, that makes sense. Thanks for the explanation!

Issue #18, opened 2016-07-07
Cielito: cannot clone from github


KineticTheory:
Cielito cannot see github.  Regressions (rtt.lanl.gov/cdash) will show 'fail' for update.  We will not fix this issue as this machine will be retired in 3 weeks.  Once Trinitite regressions are working correctly, we will stop testing on Cielito and close this ticket.

jhchang-lanl:
Cielito is no longer available.

PR #1100, opened 2021-08-06
Add autopep8 git hook

brryan:
### Background

* `flake8` is effective for enforcing python syntax, but it doesn't apply inline updates. 
* `autopep8` does provide inline fixes as an option

### Purpose of Pull Request

* Add `autopep8` git hook to apply inline formatting fixes to modified python code prior to calling `flake8`. 
* [Fixes re-git issue #18](https://re-git.lanl.gov/draco/devops/-/issues/18)

### Description of changes

* Add `pre-commit-autopep8` script and update `pre-commit` and `install-hooks.sh` to `Draco/environment/git`. 

### Status

* Reference:
* Checks
  * [ ] [Valgrind test passes](https://rtt.lanl.gov/cdash3/index.php?project=Draco)
  * [ ] [Toss3 checks pass](https://rtt.lanl.gov/cdash3/index.php?project=Draco)
  * [ ] [Trinitite checks pass](https://rtt.lanl.gov/cdash3/index.php?project=Draco)

### DIFF
--- a/environment/git/install-hooks.sh
+++ b/environment/git/install-hooks.sh
@@ -13,7 +13,7 @@
 ###########################################################
 # CONFIGURATION:
 # select which pre-commit hooks are going to be installed
-HOOKS="pre-commit pre-commit-clang-format pre-commit-flake8 pre-commit-fprettify"
+HOOKS="pre-commit pre-commit-clang-format pre-commit-autopep8 pre-commit-flake8 pre-commit-fprettify"
 HOOKS="$HOOKS pre-commit-cmake-format pre-commit-cmake-lint"
 TOOLS="common.sh"
 ###########################################################
--- a/environment/git/pre-commit
+++ b/environment/git/pre-commit
@@ -16,6 +16,8 @@
 # the order in which they are listed.
 HOOKS="pre-commit-clang-format"
 
+# only run autopep8 if the tool is available
+[[ $(which autopep8 2> /dev/null | wc -w) -gt 0 ]] && HOOKS+=" pre-commit-autopep8"
 # only run flake8 if the tool is available.
 [[ $(which flake8 2> /dev/null | wc -w) -gt 0 ]] && HOOKS+=" pre-commit-flake8"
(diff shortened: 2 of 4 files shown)


brryan:
@KineticTheory Here you go. I tested this with some simple garbled formatting in a python file and it worked for me as a commit hook.

I'm worried that at some point `autopep8` will make a change that `flake8` flags as an error, which will make committing with active hooks impossible to get past the CI... These two formatters *should* be following the same standards, but I've already encountered scenarios where `autopep8` doesn't fix things that `flake8` complains about. Maybe their options can be exactly synchronized? Another option would be to run `flake8` on each file and record the errors, then run `autopep8` to make changes, and then run `flake8` again, checking that no *new* errors were introduced by `autopep8`. A bit involved...

environment/git/install-hooks.sh
```diff
@@ -13,7 +13,7 @@
 ###########################################################
 # CONFIGURATION:
 # select which pre-commit hooks are going to be installed
-HOOKS="pre-commit pre-commit-clang-format pre-commit-flake8 pre-commit-fprettify"
+HOOKS="pre-commit pre-commit-clang-format pre-commit-autopep8 pre-commit-flake8 pre-commit-fprettify"
```
KineticTheory, line 16: Do we still need  `pre-commit-flake8` if we adopt `pre-commit-autopep8`?

KineticTheory (approved):

environment/git/install-hooks.sh
brryan: If we don't mind switching to `autopep8` from `flake8`, that would solve the above consistency issue I brought up. I do like that `autopep8` will automatically apply formatting changes, but I don't know if there are reasons for us to still prefer `flake8` at the CI level

KineticTheory: The [superlinter](https://github.com/github/super-linter) CI check that we use will run `flake8`.  I want to keep that CI because it has a very low maintenance cost.  It sounds like we need to keep both for now.

PR #1103, opened 2021-08-10
Vor2d mesh faces fix

brryan:
### Background

* The recently added python 2D voronoi generator was not outputting some variables correctly
* It's difficult to debug computational geometry without visualizing it

### Purpose of Pull Request

* [Fixes re-git issue #1350](https://re-git.lanl.gov/draco/draco/-/issues/1350)

### Description of changes

* Fix bugs in `vor_2d_mesh` mesh type with faces that share nodes and boundary data
* Add `x3d_plotter.py` script that quickly plots 2D `x3d` mesh files

### Status

* Reference:
* Checks
  * [x] Enhance `src/mesh/python` `README.md`

### DIFF
--- a/src/mesh/python/mesh_types.py
+++ b/src/mesh/python/mesh_types.py
@@ -534,14 +534,9 @@ class vor_2d_mesh(base_mesh):
             else:
                 ridge_vertices.append(verts)
 
-        # -- assign ridge vertices to regions and boundaries
+        # -- assign ridge vertices to regions
         cells = []
         cell_nodes = []
-        boundary_edges = {}
-        boundary_edges['xl'] = []
-        boundary_edges['xr'] = []
-        boundary_edges['yl'] = []
-        boundary_edges['yr'] = []
         for n, point in enumerate(points):
             cell_nodes.append(n)
             cells.append([])
@@ -561,20 +556,6 @@ class vor_2d_mesh(base_mesh):
             else:
                 # -- a boundary
                 cells[indices[0]].append(ridge_idx)
-                if (soft_equiv(vertices[v_indices[0]][0], xmin) and
-                    soft_equiv(vertices[v_indices[1]][0], xmin)):
-                    boundary_edges['xl'].append(ridge_idx)
-                elif (soft_equiv(vertices[v_indices[0]][0], xmax) and
-                      soft_equiv(vertices[v_indices[1]][0], xmax)):
-                    boundary_edges['xr'].append(ridge_idx)
-                elif (soft_equiv(vertices[v_indices[0]][1], ymin) and
-                      soft_equiv(vertices[v_indices[1]][1], ymin)):
-                    boundary_edges['yl'].append(ridge_idx)
-                elif (soft_equiv(vertices[v_indices[0]][1], ymax) and
-                      soft_equiv(vertices[v_indices[1]][1], ymax)):
-                    boundary_edges['yr'].append(ridge_idx)
-                else:
-                    assert (False), 'Boundary edge not identified'
 
         # -- update remaining base class values
         self.num_nodes = len(vertices)
@@ -591,16 +572,6 @@ class vor_2d_mesh(base_mesh):
             self.num_nodes_per_face[n] = 2
         self.faces_per_cell = cells
         self.nodes_per_face = ridge_vertices
-        self.nodes_per_side = []
-        for n in range(4):
-            bdy_key = list(boundary_edges.keys())[n]
-            bdy_nodes = []
-            for bdy in boundary_edges[bdy_key]:
-                nodes = ridge_vertices[bdy]
-                for node in nodes:
-                    if node not in bdy_nodes:
-                        bdy_nodes.append(node)
-            self.nodes_per_side.append(bdy_nodes)
 
         # -- rewrite cells and faces to not have duplicate faces
         new_cells = []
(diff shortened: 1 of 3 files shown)


brryan:
@KineticTheory were there other details you wanted added to the `README`? I'm happy to add to it

src/mesh/python/mesh_types.py
```diff
@@ -561,42 +556,72 @@ def norm(v):
...
+        for face_idx, v_indices in enumerate(self.nodes_per_face):
+            midpoint = [(vertices[v_indices[0]][0] + vertices[v_indices[1]][0]) / 2,
+                        (vertices[v_indices[0]][1] + vertices[v_indices[1]][1]) / 2]
+            distances = np.zeros(len(cell_nodes))
+            for n, node in enumerate(cell_nodes):
+                distances[n] = np.sqrt((points[n][0] - midpoint[0])**2 +
+                                       (points[n][1] - midpoint[1])**2)
+            indices = np.argsort(distances)
+            distances = np.sort(distances)
+            if not soft_equiv(distances[0], distances[1]):
```
RyanWollaeger, line 603: If true, this means that there is at least one other node closer to the mid-point of the face than the face's vertices?  It's not apparent to me how this selects for boundary faces - probably misunderstanding (sorry!).

src/mesh/python/mesh_types.py
```diff
@@ -561,42 +556,72 @@ def norm(v):
...
+        boundary_edges['xl'] = []
+        boundary_edges['xr'] = []
+        boundary_edges['yl'] = []
+        boundary_edges['yr'] = []
+        for face_idx, v_indices in enumerate(self.nodes_per_face):
+            midpoint = [(vertices[v_indices[0]][0] + vertices[v_indices[1]][0]) / 2,
+                        (vertices[v_indices[0]][1] + vertices[v_indices[1]][1]) / 2]
+            distances = np.zeros(len(cell_nodes))
+            for n, node in enumerate(cell_nodes):
+                distances[n] = np.sqrt((points[n][0] - midpoint[0])**2 +
```
RyanWollaeger, line 599: Is the square root actually needed here?

src/mesh/python/mesh_types.py
```diff
@@ -561,42 +556,72 @@ def norm(v):
...
+                      soft_equiv(vertices[v_indices[1]][1], ymax)):
+                    boundary_edges['yr'].append(face_idx)
+                else:
+                    assert (False), 'Boundary edge not identified'
         self.nodes_per_side = []
         for n in range(4):
             bdy_key = list(boundary_edges.keys())[n]
             bdy_nodes = []
             for bdy in boundary_edges[bdy_key]:
-                nodes = ridge_vertices[bdy]
```
RyanWollaeger, line 599: Why not use `np.sort`/`np.unique` to remove duplicate nodes?

RyanWollaeger (approved):
LGTM - just a few questions.

src/mesh/python/mesh_types.py
brryan: For a voronoi mesh, each cell is the region of space closest to each mesh-generating point (node). So the midpoint of each face will be equidistant to the two nodes that share that face. For boundaries, though, there's no node outside the boundary, so there will only be one closest point to the face's midpoint. Does that make sense?

src/mesh/python/mesh_types.py
brryan: It's not necessary but this isn't particularly efficient code so I left it in so `distances[n]` really is a distance. If this is bad code style though I can change it.

src/mesh/python/mesh_types.py
brryan: fixed (I think I missed where the `np.sort` should be used though)

src/mesh/python/mesh_types.py
RyanWollaeger: Oh - I see, the "nodes" are the generating points - which don't get used as the cell vertices - sorry was conflating the two - that makes sense.

src/mesh/python/mesh_types.py
RyanWollaeger: No - that is a fair point - this doesn't really need to be optimized.

src/mesh/python/mesh_types.py
brryan: sorry yeah there are a lot of terminology mismatches between scripts and files here (my fault no doubt)

RyanWollaeger: That's okay to have terminology mismatch, IMO (the hope was these classes can encapsulate the terminology the mesh developer wants to use).

src/mesh/python/mesh_types.py
RyanWollaeger: I think this question was born of my ignorance on the vertex/node terminology - is an `np.unique` actually needed here, or should the node list be unique by construction?

brryan: I'm looping over faces here when I collect all the boundary nodes/vertices and nodes/vertices here will be duplicated if two faces connect i.e. share a node/vertex. Basically I think it's just that the logic in my script is backwards from the file -- I determine boundaries via faces, but the file provides boundaries as sets of nodes/vertices.

RyanWollaeger: Ah ok - then it is indeed useful, at least in compactifying the boundary files.

PR #1114, opened 2021-08-18
Add option to generate arbitrary rectangular (prism) regions.

RyanWollaeger:
### Background

* It can be useful to delineate regions for an X3D mesh, where each region can have different properties ascribed to it.

### Purpose of Pull Request

* Add ability to specify any number of rectangular (prism) regions with arbitrary matids (called reg_ids on command line).
* For the changeset here, the `matid` field in the main X3D file is consistent with the region file cell partitioning (but the region file enumeration - Reg1, Reg2, etc - is not equal to the matid value per region).

### Description of changes

+ Add optional list inputs for region id and region bounds:
  * `--reg_ids` is a list of matids per region,
  * `--reg_bnd_per_dim` is a serialized list of boundaries dimensions, similar to `--bnd_per_dim` but for each value in `--reg_ids` (in the same order). 
+ Sanity check region list sizes if they have been specified.
+ Calculate a basic cell center and check if it is in the region.
+ Use region ids for matid field and create region file output.
+ Default to one universal region (matid=0) if regions unspecified.
+ Add example to README.md.

### Status

* Reference:
* Checks

### DIFF
--- a/src/mesh/python/x3d_generator.py
+++ b/src/mesh/python/x3d_generator.py
@@ -76,6 +76,7 @@ if (len(args.reg_ids) > 0):
     reg_dict = {reg_id: [] for reg_id in reg_ids}
     nmats = len(reg_ids)
     for cell in range(mesh.num_cells):
+        # TODO: make coordinate averaging a method for each mesh_type (permit distinct approaches)
         # -- initialize container for cell average coordinate
         coord_av = np.array([0.0, 0.0, 0.0])
         # -- average coordinates over faces first
@@ -106,9 +107,11 @@ if (len(args.reg_ids) > 0):
                 matids[cell] = reg_ids[imat]
                 reg_dict[reg_ids[imat]].append(cell + 1)
 
-    # -- remove region cells from universal region
-    for reg_id in reg_ids[1:]:
-        reg_dict[0] = [cell for cell in reg_dict[0] if cell not in reg_dict[reg_id]]
+    # -- remove region cells from universal region and lower-index regions
+    # -- this implies that this region takes precedence over regions earlier in the iteration
+    for j in range(nmats - 1):
+        for reg_id in reg_ids[j + 1:]:
+            reg_dict[j] = [cell for cell in reg_dict[j] if cell not in reg_dict[reg_id]]
 
     # -- convert dictionary of lists to dictionary of numpy arrays
     for reg_id in reg_ids:
--- a/src/mesh/python/x3d_generator.py
+++ b/src/mesh/python/x3d_generator.py
@@ -22,11 +22,15 @@ parser.add_argument('-mt', '--mesh_type', type=str, default='orth_2d_mesh',
 parser.add_argument('-nd', '--num_per_dim', type=int, nargs='+', default=[1, 1],
                     help='Number of cells per dimension.')
 parser.add_argument('-bd', '--bnd_per_dim', type=float, nargs='+', default=[0.0, 1.0, 0.0, 1.0],
-                    help='Length per dimension.')
+                    help='Bounds per dimension.')
 parser.add_argument('--name', type=str, default='mesh',
                     help='Select file name (will be prefixed with x3d. and sufficed with .in).')
 parser.add_argument('--num_cells', type=int, default=100,
                     help='Number of cells')
+# -- optional region/matid input
+parser.add_argument('--reg_ids', type=int, nargs='+', default=[], help='List of region ids.')
+parser.add_argument('--reg_bnd_per_dim', type=float, nargs='+', default=[],
+                    help='Bounds per dimension of region.')
 
 # -- parse arguments from command line
 args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


src/mesh/python/x3d_generator.py
```diff
@@ -51,6 +59,65 @@
...
+    reg_ids = args.reg_ids
+    reg_bnd_per_dim = args.reg_bnd_per_dim
+    reg_ids = [0] + reg_ids
+    reg_bnd_per_dim = args.bnd_per_dim + reg_bnd_per_dim
+    # -- initialize dictionary of cell lists
+    reg_dict = {reg_id: [] for reg_id in reg_ids}
+    nmats = len(reg_ids)
+    for cell in range(mesh.num_cells):
+        # -- initialize container for cell average coordinate
+        coord_av = np.array([0.0, 0.0, 0.0])
```
RyanWollaeger, line 81: This cell-average coordinate computation should probably be a method for each mesh (for Voronoi it would just be the original sampled points used to generate the mesh).  However, this would be a fairly minor change (and just a refactor, not a significant feature change).

KineticTheory (approved):
LGTM - but I just did a cursory review.  I'll leave it to @brryan to check that the update works as advertised.

src/mesh/python/x3d_generator.py
```diff
@@ -51,6 +59,65 @@
...
+            # -- get the node coordinates for this face to average
+            for node in mesh.nodes_per_face[face]:
+                for idim in range(mesh.ndim):
+                    coord_face_av[idim] += mesh.coordinates_per_node[node, idim]
+            # -- finish face average by dividing by number of nodes per face
+            coord_face_av /= mesh.num_nodes_per_face[face]
+            # -- add face average to cell average
+            coord_av += coord_face_av
+        # -- finish cell average by dividing by number of faces per cell
+        coord_av /= mesh.num_faces_per_cell[cell]
```
brryan, line 95: Will this work for all concave cells?

brryan (commented):
Just one question on whether we are specializing to more-or-less convex polygons for cells here, otherwise LGTM

src/mesh/python/x3d_generator.py
RyanWollaeger: Good question - there are sort of a few answers to this.  The coordinate average here:
1.) can be outside of cell if the cell is concave,
2.) serves as an indicator for whether the cell is included in a particular region,
3.) but the region is merely an overly (e.g painting a label), so doesn't affect the mesh structure,
4.) cells can be included in a region without being entirely inside it, so there is error on the order of the cell resolution.

RyanWollaeger: Putting (1)-(4) together, if a cell is extremly concave, it is possible (pretty sure at least) most of it's volume will be out of the region, while the `coord_av` will be in the region.  This is why I had commented that this averaging should be a method in the `mesh_type` mesh classes (but it didn't seem urgent to do so).  Should I go ahead and make this change here (it is not too difficult)?

brryan: > This is why I had commented that this averaging should be a method in the mesh_type mesh classes

Ah OK I misunderstood that comment above. 

I think this is fine as is for now (especially since, like you say, this is really a spatial error of a type that is already inherent in the mesh scheme), but what do you think about adding a comment somewhere here e.g. `TODO(RW) make coordinate averaging a separate method for each mesh_type to account for concave cells`? (That's 100 characters so you might have a better comment in mind)

src/mesh/python/x3d_generator.py
```diff
@@ -51,6 +59,65 @@
...
+# -- set matid data (using region ids and bounds if available)
+matids = np.zeros(mesh.num_cells, dtype=int)
+reg_dict = {}
+reg_ids = []
+if (len(args.reg_ids) > 0):
+    # -- prepend reg_ids and reg_bnd_per_dim with universal region
+    reg_ids = args.reg_ids
+    reg_bnd_per_dim = args.reg_bnd_per_dim
+    reg_ids = [0] + reg_ids
+    reg_bnd_per_dim = args.bnd_per_dim + reg_bnd_per_dim
```
brryan, line 74: Is it worth checking here that a universal region wasn't passed in through the command line args by an unsuspecting user?

jhchang-lanl (approved):

src/mesh/python/x3d_generator.py
RyanWollaeger: Good idea! Will add a check for this...

RyanWollaeger: This may be okay, since the universal region does not get added if all its cells are covered by 1 or more other regions.  However, removing cells from the universal region (originally at line 111) that overlap with other regions was only being done for the universal region, and not for specified regions, which could cause overlap.  To avoid this, in the recent commit line 111 has been generalized to subtract cells from regions overlapping with regions specified later in the user input list (this has been tested on the 2-region example that was added to the README).

src/mesh/python/x3d_generator.py
RyanWollaeger: Just added a TODO comment for this.

RyanWollaeger:
It appears one coverage check is not passing.  Going to try rebasing this PR against develop to see if it is avoidable.

KineticTheory:
> 
> 
> It appears one coverage check is not passing. Going to try rebasing this PR against develop to see if it is avoidable.

Don't bother with the rebase. The coverage is still good. If this is ready, let's just merge it.

RyanWollaeger:
> > It appears one coverage check is not passing. Going to try rebasing this PR against develop to see if it is avoidable.
> 
> Don't bother with the rebase. The coverage is still good. If this is ready, let's just merge it.

Sorry - missed your comment and just rebased (doesn't hurt I suppose).

brryan:
LGTM

